PR #3, opened 2017-11-19
Reduce dependencies

ChrisRackauckas:


ChrisRackauckas:
Instead of grabbing DifferentialEquations.jl as a whole, it only depends on DiffEqBase.jl now. Then the tests grab ParameterizedFunctions.jl

korsbo:
That's indeed better, thanks!

PR #7, opened 2018-02-02
Remove unnecessary backslash in user defined functions

goropikari:
If an expression contains a user defined function such as f(x), `latexify` generates LaTeX command with unnecessary backslash, `\f(x)`.
```julia
latexify("f(x)") # --> L"$\f\left( x \right)$"
latexraw("f(x)") # --> "\\f\\left( x \\right)"
```
Of course, those backslashes are useful for trigonometric functions and Greek letters.
So I have modified `latexoperation` to resolve the above problem.


korsbo:
This is excellent, thank you!

PR #25, opened 2018-07-15
Major

korsbo:
Reduce the package to a single function, `latexify()`. The output format can still be specified by the keyword `env=`. 

Bring the documentation up to speed (finally!).

This should ready the package for a fairly major release.


korsbo:
Bring the documentation up to speed for a new release where `latexify(x; env=)` has taken over all other functions.

PR #29, opened 2018-08-09
Julia07

korsbo:
Get everything up an running on Julia 0.7 (and hopefully 1.0).


korsbo:
Latexify now works fine, but Documenter.jl is giving me grief. If I were to merge this branch, the package would work with 0.7 and 1.0, but the documentation would not be generated correctly.

Issue #33, opened 2018-09-17
Support for broadcasting?


jlperla:
```
using Latexify
ex = :(exp.(z)
)
latexify(ex)
```
Gives 
```
Latexify.jl's latexoperation does not know what to do with one of the
                operators in your expression (z).
```

Of course, this works great with `:(exp(z))`.  My suggestion: Latexify just strips out all broadcasting in a pre-step?

jlperla:
I guess it depends on what the intention is of the library.  For me checking algebra, dropping the `.` everywhere makes things **a lot** clearer.  Consider whether this should be an option in the `latexify` function?  `latexify(ex; debroadcast = false)`?

For the `dot` ignoring list, I would also add in `>, <, >=, <=, `.  I ran into an issue with
```julia
using Latexify
latexify(:(2 * (a .<= b)))
```
Where the `.<=` became pre-fix instead of infix.  I would rather just see this become `2 \cdot (a \leq b)`

PR #36, opened 2018-10-12
Update support for broadcasted operations. 

korsbo:
See #33 

This strips all dots and basically ignores broadcasting.

I cringe a bit as a mathematician since this allows for misrepresentations without raising errors. However, I agree that this is more useful than an error, and I will defer the correctness of the mathematics to the user. 

This also adds infix support for (chained) comparison operations:
```julia
latexify(:(a .<= b .< c .== a))
```
$\left( a \leq b \lt c = a \right)$

### DIFF
--- a/src/latexoperation.jl
+++ b/src/latexoperation.jl
@@ -58,8 +58,41 @@ function latexoperation(ex::Expr, prevOp::AbstractArray)
 
     if ex.head == :.
         ex.head = :call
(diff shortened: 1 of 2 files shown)


korsbo:
In order to determine if I really need parentheses I would either have to do something very hacky, or dig deep into the code. For now, I just enforce parentheses all the time.

Issue #32, opened 2018-08-18
Writing Float64 in scientific notation


elchiconuevo:
This is a fantastic package thanks to which I'm able to export data from Julia onto my .tex document as a table. I'm having some minor issues with this feature. I'd want to convert the ugly numbers (`12434354.23234`) into a scientific format (`1.24e+7`) which is usually preferred for tables in LaTeX. Is it possible to do this at the moment?

korsbo:
Thanks for your kind words, I'm happy to have been of some use.

The feature you are asking for does not exist although it absolutely should. I'm just now trying to design a neat way of implementing it. If you cant wait, it is easy to just overload `latexraw` like this:

```julia
using Latexify
using Printf # Only needed in Julia 0.7 or higher
import Latexify; latexraw

latexraw(x::Real) = @sprintf("%.2e", x)
```
and then you should get:

```julia
latexify( 1e6 .* rand(3,3) )
```
```math
\begin{equation}
\left[
\begin{array}{ccc}
3.96e+05 & 2.74e+05 & 2.21e+04 \\ 
9.00e+05 & 7.61e+04 & 7.87e+05 \\ 
2.02e+05 & 9.74e+05 & 8.92e+05 \\ 
\end{array}
\right]
\end{equation}
```

I haven't quite figured out how best to implement this with keyword args. 
```julia
latexraw(x::Real; format="%.2e") = @sprintf(format, x)
```
does not work. I'll have to think of something else.

korsbo:
I have found a nice way of doing it, but it will require a little bit more work than I have time for today. I'll get to it soon. Overloading `latexraw` should work in the meantime.

elchiconuevo:
Thanks for the reply. In the meanwhile, I just converted my array of type `Float64` into `String` using the `@sprintf` command and it works for now. Will give `latexraw` a try.

korsbo:
This feature was added in #39.

PR #39, opened 2018-12-10
Formatted numbers

korsbo:
As per issue #32.

Add a kwarg `fmt` which is passed along to Printf for number formatting.  This enables for example scientific notation to be used.


korsbo:
closes #32

PR #47, opened 2019-01-30
Add escape_underscore kwarg for Markdown.

korsbo:
Underscores can denote italics in markdown. Sometimes this is not what you want. This adds a kwarg to escape the underscores.


korsbo:
The failure is on Nightly and it is due to precompilation errors of other packages. This PR is not responsible for them. Merging.

PR #53, opened 2019-03-22
add usecdot kwarg

isaacsas:
This adds a `usecdot` kwarg to `latexify`. The default is true, which should not result in any changes in generated Latex, but if set to `false` will cause `latexraw` to just insert a single space within products instead of a \cdot.

I tried to forward kwargs everywhere it looked like this would change behavior, but please do let me know if I've missed anything and I can update the PR.

### DIFF
--- a/docs/src/table_generator.jl
+++ b/docs/src/table_generator.jl
@@ -36,6 +36,7 @@ keyword_arguments = [
     KeywordArgument(:fmt, [:mdtable, :tabular, :align, :array, :raw, :inline], "format string", "`\"\"`", "Format number output in accordence with Printf. Example: \"%.2e\"", [:Any]),
     KeywordArgument(:escape_underscores, [:mdtable, :mdtext], "`Bool`", "`false`", "Prevent underscores from being interpreted as formatting.", [:Any]),
     KeywordArgument(:convert_unicode, [:mdtable, :tabular, :align, :array, :raw, :inline], "`Bool`", "`true`", "Convert unicode characters to latex commands, for example `α` to `\\alpha`", [:Any]),
+    KeywordArgument(:cdot, [:mdtable, :tabular, :align, :array, :raw, :inline], "`Bool`", "`true`", "Toggle between using `\cdot` or just a space to represent multiplication.", [:Any]),
 #     KeywordArgument(:template, [:array], "`Bool`", "`false`", "description", [:Any]),
     ]
 
--- a/src/latexoperation.jl
+++ b/src/latexoperation.jl
@@ -7,7 +7,7 @@ This uses the information about the previous operations to decide if
 a parenthesis is needed.
 
 """
-function latexoperation(ex::Expr, prevOp::AbstractArray; usecdot=true, kwargs...)
+function latexoperation(ex::Expr, prevOp::AbstractArray; cdot=true, kwargs...)
     op = ex.args[1]
     convertSubscript!(ex)
     args = ex.args
@@ -20,7 +20,7 @@ function latexoperation(ex::Expr, prevOp::AbstractArray; usecdot=true, kwargs...
             arg = args[i]
             prevOp[i] in [:+, :-]  && (arg = "\\left( $arg \\right)")
             str = string(str, arg)
-            i != length(args) && (str *= usecdot ? " \\cdot " : " ")
+            i != length(args) && (str *= cdot ? " \\cdot " : " ")
         end
         return str
 
--- a/src/latexoperation.jl
+++ b/src/latexoperation.jl
@@ -7,7 +7,7 @@ This uses the information about the previous operations to decide if
 a parenthesis is needed.
 
 """
-function latexoperation(ex::Expr, prevOp::AbstractArray)
+function latexoperation(ex::Expr, prevOp::AbstractArray; usecdot=true, kwargs...)
     op = ex.args[1]
     convertSubscript!(ex)
     args = ex.args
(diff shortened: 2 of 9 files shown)


isaacsas:
@korsbo Just to add, if this isn't of interest feel free to close. I was finding for systems like the repressilator the cdots caused the generated ODEs, and particularly the Jacobian matrix, to be harder to read. But that's certainly just a personal preference...

korsbo:
I'm very happy to merge this after we add a few things.

### Fix function ambiguity?

First, should we make sure that `f * (x + y)` always have a `\cdot` in order to distinguish it from a function `f(x + y)`? I'm open to hearing arguments against it, but this would make sense to me.

Implementing this would require some tinkering around https://github.com/korsbo/Latexify.jl/blob/acc8fbf1f37563a990a1c4be0b2d7efda0aec025/src/latexoperation.jl#L21
It should not be too hard, but it requires more brainpower and time than my newborn twins leaving are me at the moment. 

### Documentation
The argument tables of the documentation are generated automatically from `test/src/table_generator.jl`. Adding a line like 

```julia
    KeywordArgument(:cdot, [:mdtable, :tabular, :align, :array, :raw, :inline], "`Bool`", "`true`", "Toggle between using `\cdot` or just a space to represent multiplication.", [:Any]),
```
to the [keyword arguments list](https://github.com/korsbo/Latexify.jl/blob/acc8fbf1f37563a990a1c4be0b2d7efda0aec025/docs/src/table_generator.jl#L23) would take care of that. 

### Testing 
It would be simple to add some tests for the different environments with both `cdot=true` and `cdot=false`. I have a macro which automatically generates tests from a function and copies the result to your clipboard: 

```julia
using Latexify
Latexify.@generate_test latexify( :(x*y); env=:inline, cdot=false)
```
places 
```
@test latexify("x*y"; env=:inline) == 
       raw"$x \cdot y$"
```
in your clipboard. Test generation becomes pretty quick with this. 

This is a good addition, thanks Sam!

isaacsas:
The documentation and testing suggestions make sense, and I am happy to add them (hopefully this weekend when I have some free time). I didn't want to do anything along those lines till hearing if you were interested in this.

Regarding distinguishing between `f*(x+y)` and `f(x+y)`; I don't think in papers I usually explicitly use a `*`, but instead assume the reader will understand what I mean from the symbol's definition (i.e. if `f` is a function usually the default is to assume function evaluation, whereas if `f` is a scalar this means multiplication). I wouldn't usually write `f*g*(x+y)`, but instead use explicit arguments and/or brackets, i.e. `f(x)g(x)(x+y)` or `f(x)g(x)[x+y]`. In any case, I will think a bit about how to handle this, but it may require more expression manipulation knowledge than I currently have.

isaacsas:
Ahh, I think I misunderstood your suggestion. It wasn't about detecting functions, but about multiplying a term already in parenthesis. I still would usually let context and symbol definitions handle the difference when writing Latex, but that seems more doable, so I can look into it.

isaacsas:
So I tried a few different simple rules, such as:
```julia
if prevOp[i] in [:+, :-] 
    arg = (usecdot || prevOp[i-1] ∈ [:+, :-]) ? " \\left( $arg \\right)" : " \\cdot \\left( $arg \\right)"
end
str = string(str, arg)
i != length(args) && (str *= usecdot ? " \\cdot " : " ")
```
When `usecdot=false`, this gives for 
```julia
latexify( :(f*(k+m)*g*(g+y)*(h+x)*k), usecdot=false)
```
the latex
```
L"$f  \cdot \left( k + m \right) g  \cdot \left( g + y \right)  \left( h + x \right) k$"
```
At this point, I think having a mix of cdots and no cdots looks unappealing, and if I was type setting this I would probably just include cdots throughout or add parenthesis around the cdot'ed pairs. My feeling is it is better to just have `usecdot=false` drop all cdots. Since the default will be to always include them, a user will have to knowingly opt out of having them (and hopefully realize how this changes the generated expression). Another reason to not include them before parenthesis is that I'd prefer an expression like `:(α*(x+y))` to always give `\alpha \left(x +y\right)`, since I almost always mean \alpha to be a constant. There isn't any way to distinguish this from `:(f*(x+y))` though (short of making some type of lookup table of common function names). 

Anyways if you have other suggestions about this let me know, otherwise my inclination is to leave the behavior as in the original commit (modulo changing the kwarg to `cdot` like you suggested).

isaacsas:
Ok, I've added tests and updated table_generator.jl as you suggested. Let me know if you need anything else for it to be ok for merging (assuming tests pass).

isaacsas:
The .7 test failures seem to be coming from an old DiffEqBiological version (we haven't supported pre Julia 1.0 in a while). Is it still necessary to test against .7?

isaacsas:
If you do want to keep testing against .7 I can comment out the reaction_network tests till you drop that dependency. Just let me know how you'd prefer I proceed.

korsbo:
I apologise for being a bit slow in our correspondence, my PhD and newborn twins do not leave me much time for my other projects. 

I don't mind dropping support for 0.7 when it is actually needed. But so far, latexify works fine for 0.7, it's just the use of DiffEqBiological.jl in the test which is tripping things up. What we could do is to limit that test to Julia 1.0 or above using something like:

```julia
if VERSION >= v"1.0"
  using DiffEqBiological
  ...
end
```

Apart from this, I'm happy with it all. As you say, we can skip the `\cdot` even in the cases where it could be confusing and just leave that up to the user.

isaacsas:
No worries, there's no rush on this! I'll add the version check later today. Thanks!

isaacsas:
One caveat, I didn't indent the code within the `if VERSION >=` check as it led to tests failing (since the raw strings then include the spaces for indentation).

isaacsas:
Ok, should be good now.

korsbo:
Great, thanks!

Issue #58, opened 2019-04-23
Fix table generation in documents. 


korsbo:
The tables of possible arguments are broken for the latest docs. Fix before release.

korsbo:
This should be fixed now.

PR #59, opened 2019-06-26
Add support for env=:equation.

korsbo:
When using Weave.jl, I found myself missing the option of outputting a 
```latex
\begin{equation}
...
\end{equation}
```
environment. This PR remedies this.

- [x] Add the code for this. 
- [x] Update the docs.
- [x] Add  tests.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -3,7 +3,7 @@ using Latexify
 
 makedocs(
     modules = [Latexify],
-    format = :html,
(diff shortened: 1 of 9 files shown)


korsbo:
This also fixes a syntax error in the doc generation. Closes #58.

korsbo:
Some unrelated tests are failing. They have nothing to do with this PR, so I will merge this and see if I can fix the broken tests in a separate PR.

PR #64, opened 2019-07-15
Add support for ModelingToolkit.jl

korsbo:
- [x] code
- [ ] tests
- [ ] documentation

It is mostly functional already.

```julia
using Latexify
using ModelingToolkit

@parameters t σ ρ β
@variables x(t) y(t) z(t)
@derivatives D'~t

eqs = [D(x) ~ σ*(y-x)*D(x-y)/D(z),
       0 ~ σ*x*(ρ-z)/10-y,
       D(z) ~ x*y - β*z]

latexify(eqs; iv=:t)
```

$$
\begin{align}
\frac{dx(t)}{dt} =& \frac{\sigma \cdot \left( \mathrm{y}\left( t \right) - \mathrm{x}\left( t \right) \right) \cdot \frac{d\left(\mathrm{x}\left( t \right) - \mathrm{y}\left( t \right)\right)}{dt}}{\frac{dz(t)}{dt}} \\
0 =& \frac{\sigma \cdot \mathrm{x}\left( t \right) \cdot \left( \rho - \mathrm{z}\left( t \right) \right)}{10} - \mathrm{y}\left( t \right) \\
\frac{dz(t)}{dt} =& \mathrm{x}\left( t \right) \cdot \mathrm{y}\left( t \right) - \beta \cdot \mathrm{z}\left( t \right)
\end{align}
$$

image

### Remaining problems
- I'm letting the user supply a single independent variable as a kwarg. This will not work with PDEs and it's less elegant than I'd like.
- The output is not consistent with that of `@ode_def` or `@reaction_network`.
  - variables are surrounded by `\textrm{}` (I don't mind but I dislike inconsistency).
  - the independent variable is shown `x(t)`. 
  - brackets are not supported.
- Variables are in italics when they are in derivatives but not otherwise.
- ModelingToolkit represents the derivative in two different ways for my two different environments (Julia 1.1 and 1.2-rc2) only one of them works correctly. I'm guessing that this is a problem with ModelingToolkit and I don't want to hack together a fix here. 1.1 gives something like `Equation((D'~t())(z(t()))` while 1.2 gives `Equation(Differential(z(t)))`.

### DIFF
--- a/src/Latexify.jl
+++ b/src/Latexify.jl
@@ -46,6 +46,9 @@ function __init__()
     @require DiffEqBiological = "eb300fae-53e8-50a0-950c-e21f52c2b7e0" begin
         include(joinpath("plugins", "DiffEqBiological.jl"))
     end
+    @require ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78" begin
+        include(joinpath("plugins", "ModelingToolkit.jl"))
+    end
     @require SymEngine = "123dc426-2d89-5057-bbad-38513e3affd8" begin
         include(joinpath("plugins", "SymEngine.jl"))
     end
--- a/src/plugins/ModelingToolkit.jl
+++ b/src/plugins/ModelingToolkit.jl
@@ -1,0 +1,30 @@
+
+##################################################
+#   Override default handling (default = inline) #
+##################################################
+get_latex_function(eqs::Vector{ModelingToolkit.Equation}) = latexalign
+
+get_md_function(eqs::Vector{ModelingToolkit.Equation}) = mdalign
+
+###############################################
+#         Overload environment functions      #
+###############################################
+
+function latexalign(eqs::Vector{ModelingToolkit.Equation}; iw=:t, kwargs...)
+    rhs = getfield.(eqs, :rhs)
+    rhs = convert.(Expr, rhs)
+    rhs = [postwalk(x -> x isa ModelingToolkit.Constant ? x.value : x, eq) for eq in rhs]
+    rhs = [postwalk(x -> x isa Expr && length(x.args) == 1 ? x.args[1] : x, eq) for eq in rhs]
+    rhs = [postwalk(x -> x isa Expr && x.args[1] == :Differential && length(x.args[2].args) == 2 ? :($(Symbol(:d, x.args[2]))/($(Symbol(:d, x.args[2].args[2])))) : x, eq) for eq in rhs]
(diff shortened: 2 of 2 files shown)


korsbo:
@ChrisRackauckas

ChrisRackauckas:
>ModelingToolkit represents the derivative in two different ways for my two different environments (Julia 1.1 and 1.2-rc2) only one of them works correctly. I'm guessing that this is a problem with ModelingToolkit and I don't want to hack together a fix here. 1.1 gives something like Equation((D'~t())(z(t())) while 1.2 gives Equation(Differential(z(t))).

Same version of MT?

>I'm letting the user supply a single independent variable as a kwarg. This will not work with PDEs and it's less elegant than I'd like.

What about doing it on the `ODESystem` instead?

korsbo:
> Same version of MT?

yes, 0.5.0. I can file an issue if you'd like.

> What about doing it on the `ODESystem` instead?

I have thought about it and it would solve the problems of identifying both the independent and the dependent variables (for brackets). On the other hand, latexify is more forgiving than `ODESystem()`, if you pass the example above to `ODESystem` you will get an error but latexify does not mind. Still, if latexify can handle any `AbstractSystem` then it should be usable on any equations which are actually capable of being simulated so that may be the best option.

ChrisRackauckas:
>yes, 0.5.0. I can file an issue if you'd like.

Yes please.

>Still, if latexify can handle any AbstractSystem then it should be usable on any equations which are actually capable of being simulated so that may be the best option.

I think arbitrary equation support is good, if it can fully work, but dispatches on systems would be nice and carry a bit more information.

korsbo:
Let's just add everything. We could have arbitrary equation support which either takes the variables and stuff as kwargs or where I try to dig out the code in ModelingToolkit that infers them. Methods for `AbstractSystems` could then just call the method for `Vector{Expression}` and automatically specify those kwargs.

I'm trying to submit a paper this week though (I will likely fail but nevertheless) so this will have to wait a bit.

korsbo:
This new latexify recipe system actually makes it more sensible to define ModelingToolkit support in ModelingToolkit itself. A lot of the code in this PR can however be copied.

ChrisRackauckas:
Yes indeed! A recipe sounds like the right way to do it. Could you at least give the starter on that?

korsbo:
> Could you at least give the starter on that?

Done! https://github.com/JuliaDiffEq/ModelingToolkit.jl/pull/190

korsbo:
I have cannibalised all the relevant code from this PR to JuliaDiffEq/ModelingToolkit.jl#190. The discussion could still be relevant, but I see no need to keep this PR open.

PR #69, opened 2019-09-10
Add ability to set default keyword arguments.

### LINKED ISSUE #60: Add ability to set default kwargs.
One often uses the same set of kwargs for an entire session. It would make sense to be able to change the default kwarg value with a single call. 

- Add a dict of `:kwarg => default_value`.
- Update all kwargs defaults to fetch the value from the dict.
- Add a function to update the values in the dict. 
- Update docs (let the table generator fetch default values from the dict).

korsbo:
closes #60

### DIFF
--- a/test/latexify_test.jl
+++ b/test/latexify_test.jl
@@ -28,17 +28,20 @@ raw"$x \cdot y$"
 
 set_default(cdot = false)
 
-Latexify.@generate_test latexify("x * y")
-
 @test latexify("x * y") == 
 raw"$x y$"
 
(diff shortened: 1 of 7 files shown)


test/latexify_test.jl
```diff
@@ -20,3 +20,27 @@ raw"\begin{align*}
...
 "
+ 
+#### Test the setting of default keyword arguments.
+
+@test latexify("x * y") == 
+raw"$x \cdot y$"
+
+set_default(cdot = false)
+
+Latexify.@generate_test latexify("x * y")
```
korsbo: This should not have been left behind.

Issue #67, opened 2019-07-24
latexify returning row instead of column


philschrad:
From 
```
B=[1 0 5; 2 5 4]
C=[2;3;1]
latexify(B*C)|>print
```
I get
```
\begin{equation}
\left[
\begin{array}{cc}
7 & 23 \\
\end{array}
\right]
\end{equation}
```
but I think it should be 
```
\begin{equation}
\left[
\begin{array}{cc}
7 \\ 23 
\end{array}
\right]
\end{equation}
```
p.s. great package, thanks!

korsbo:
Both of the behaviours are supported. You can specify `transpose=true` to get what you want. However, it does indeed seem a bit inelegant to have the default behaviour deviate from standard mathematical notation, thanks for reporting! 

The reason I did this in the first place was that I used latexify to visualise vectors while coding in jupyter and horizontal vectors are less imposing as they take up less space. With #60, this will become less of an issue. 

I agree that the default should be changed and it would be trivial to do so. However, the change would be breaking so I'll postpone it to the next major release.

korsbo:
I can see how that would be a problem. I'm fast-tracking this to the next minor release. [SemVer](https://semver.org/) allows for this (even though it is breaking) since latexify is still at `v0.x`. 

The coding is already done and any desperate soul can just use the `fix_transpose` branch until the next version is released. 
```julia
] add Latexify#fix_transpose
```

Just make sure to free Latexify.jl after the change has passed through as a release.
```julia
] free Latexify
```

korsbo:
This is now merged with master and will be released as v0.11.0 within a day or so.

lukeburns:
🙌

PR #70, opened 2019-09-13
Redefine default orientation vectors in env=:array

korsbo:
See #67. 

This is a breaking change and it affects how vectors and tuples are represented in the `:array` environment.


korsbo:
Note that this PR is not for master but for the minor branch. I want master to be accessible for hotfixes between minor releases.

PR #80, opened 2019-10-28
Fix broken mid-word unicode. Closes #79.

### LINKED ISSUE #79: Latexify outputs incorrectly when variable names contain both unicode and ascii
[Discourse links](https://discourse.julialang.org/t/latexify-jl-outputs-incorrectly-when-variable-names-contain-both-unicode-and-ascii/30393)
It seems that if a variable contains both unicode and ascii character, latexify will product incorrect output:
```
julia> latexify("δa")
L"$\deltaa$"
```
but what I want is :`$\delta a$,` adding `_` would not help because it will change into `\delta_{a}`。

I am not sure whether this is intended.  But since it's impossible to use spaces in variable names, so I can't get around it.

korsbo:
```
julia> latexify(:αβa)                                                                  
L"$\alpha{\beta}a$" 
```

This avoids the error `\betaa`

### DIFF
--- a/src/unicode2latex.jl
+++ b/src/unicode2latex.jl
@@ -2452,15 +2452,20 @@ const unicodedict = Dict{Char, String}(
     '•' => raw"\bullet",
     )
 
-
 function unicode2latex(str::String)
     isascii(str) && return str
-    str = join(
-        map(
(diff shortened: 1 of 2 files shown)


korsbo:
Why has this not triggered Travis?

PR #95, opened 2020-04-28
Add `rows` kwarg to latexalign.

korsbo:
Allows the specification of which rows to output in an align.


korsbo:
@ChrisRackauckas, we chatted about something like this a little while ago so I thought I'd ping you.

ChrisRackauckas:
Oh that's cool! Make sure it gets in the docs!

PR #96, opened 2020-05-07
Ensure that repeated subscripts do not cause invalid latex

asinghvi17:
This was a problem in ModelingToolkit - a variable `n₁₂` latexified to `n_1_2`, which is a problem.

### DIFF
--- a/src/unicode2latex.jl
+++ b/src/unicode2latex.jl
@@ -2459,7 +2459,7 @@ function unicode2latex(str::String)
 
     for (i, char) in enumerate(str)
         if str_array[i] isa String
-            if i < str_length && str_array[i+1] isa Char && (isletter(str_array[i+1]) || isdigit(str_array[i+1]))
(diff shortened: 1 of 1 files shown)


asinghvi17:
Ah, looking deeper into the logic this won't fix it.  I'll change the Unicode map directly for the time being.

thchr:
The current situation seems quite broken for non-solitary sub/superscripts in general, e.g.:
```jl
julia> latexify("xᵃᶜ")
L"$x^a^c$"

julia> latexify("xₐᵦ")
L"$x_{a\_beta}$"

julia> latexify("xₐe")
L"$x{_{a}e}$"

julia> latexify("xₐₚ")
L"$x_{a\_p}$"

julia> latexify("1¹²")
ERROR: in Latexify.jl: 
You are trying to create latex-maths from a `String` that cannot be parsed as
an expression. 
...
```

korsbo:
This should be fixed for most use-cases in #121.

PR #97, opened 2020-05-14
Add booktabs support

EuklidAlexandria:
It now can automatically insert the booktabs-package commands `\toprule`, `\midrule` and `\bottomrule`.


korsbo:
Nice, thanks!

korsbo:
This is just awaiting merging to the general registry. Should not take more than an hour or so.

Issue #28, opened 2018-07-18
Add support for native Julia functions?


korsbo:
In Julia, one can inspect the expressions that build up a method with `@code_lowered`. The returned object does have all the information required to deduce what the method does. I have been playing around with latexifying simple julia functions. The code for this can be found in the `function_support` branch.

I have implemented rudimentary support for latexifying some simple functions:
```julia
using Latexify
gamma_pdf(t, n, r) = r^n*t^(n-1)*exp(-r*t)/gamma(n)

latexify(gamma_pdf, (1,1,1))
```
$\frac{r^{n} \cdot t^{n - 1} \cdot e^{ - r \cdot t}}{\Gamma\left( n \right)}$

(I need to pass a tuple or array of arguments to my `gamma_pdf` so that Julia can figure out which method to use)

A multi-line example:

```julia
function ode(du, u, p, t)
    du[1] = p[1] * u[2] 
    du[2] = - u[1]
end
latexify(ode, (1,1,1,1))
```
\begin{equation}
\begin{array}{l}
\textrm{du}\left[1\right] = p\left[1\right] \cdot u\left[2\right] \\\\
\textrm{du}\left[2\right] =  - u\left[1\right] \\\\
\end{array}
\end{equation}

However, this is a bit of a house of cards in two repects: One is that the code is hard to read/understand, which makes the package harder to maintain. The other respect is that It breaks for any operation that is not explicitly defined in `latexoperation`. 

For example:
```julia
f(n) = [i for i in 1:n]
latexify(f, (1))
```
will error.

Support for very simple stuff is achievable. Support for truly general functions should be theoretically possible, but I can't see it being feasible. 

Does anyone have feedback on this/is anyone willing and able to implement a better version than I have done myself?

Also, is this worth working further on at all? It has already been a bit of a time-sink.

korsbo:
I'm dropping this for the foreseeable future. The code that I once wrote was for Julia 0.6 and is no longer functional. This would take a long time and it would most likely never be as general as one would like.

korsbo:
The following works for later Julia versions.

```julia
f(x, y, z) = x * x / (y + z)
c = @code_lowered f(1., 1., 1.)
exparr = [postwalk(x -> x isa Core.SlotNumber ? c.slotnames[x.id] : x, expr) for expr in c.code]

expand_ssa(ex, exparr) = postwalk(x -> x isa Core.SSAValue ? exparr[x.id] : x, ex)

[...]

PR #100, opened 2020-06-04
Add macro for latexifying one-liner functions and add external rendering.

korsbo:
Poor practice, but this PR contains two separate new features.

One is the exported function `render` which tries to render a LaTeXString using lualatex and open the result in an external viewer. Useful for rendering things when in the REPL.

The other is a macro `@latexify` which can be used to directly latexify one-liner Julia functions.
```julia
julia> @latexify f(x, y=2) = x^2 + y
$\mathrm{f}\left( x, y = 2 \right) = x^{2} + y$
```
This macro call will also define `f` as expected.

No documentation or test yet.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,7 @@ name = "Latexify"
 uuid = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
 authors = ["Niklas Korsbo <niklas.korsbo@gmail.com>"]
 repo = "https://github.com/korsbo/Latexify.jl.git"
-version = "0.13.1"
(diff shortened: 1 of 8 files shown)


korsbo:
I disabled execution from `@latexify`. This macro now just treats the argument as an expression for latexification, without executing that expression.

The new macro `@latexrun` will first execute the expression and then latexify it.

Both should work for some simple maths-looking functions like `f(x; y=1) = x + y`.

korsbo:
Although the approach is different than I imagined, I would say that this PR closes #28.

PR #105, opened 2020-06-26
Use standard syntax for including files

MikeInnes:
There's no reason to use `joinpath` here since this syntax is supported across different OSs, and it has the potential to cause problems with tools that try to statically understand `include`s.


ViralBShah:
I believe this will make it easier to build Latexify into system images.

korsbo:
Ok, thanks for enlightening me 🙂

PR #109, opened 2020-07-23
Make latexraw recursive for Rational and Complex args

gustaphe:
`latexraw(x::Rational{T<:Integer})` and `latexraw(z::Complex{T<:Real})` now recursively use `latexraw(x::T)`, so that for instance `latexify( Complex(1//3,1//2) )` evaluates to `L"$\frac{1}{3}+\frac{1}{2}\textit{i}$"`.


korsbo:
That makes perfect sense. Thanks!

PR #113, opened 2020-08-21
Support {cases} environments and logical operators

mossr:
Pardon four changes bundled into one PR. The main contribution is adding support to translate Julia conditional blocks (and ternary ifs) to LaTeX `cases` environments. See an example with stressing cases here:

- PDF: https://github.com/mossr/TeX.jl/blob/master/test/pdf/latexify_cases.pdf
- Code: https://github.com/mossr/TeX.jl/blob/master/test/test_latexify_cases.jl

The other minor contributions are:

1. Add support for logical operators: `\wedge` for `&&`, `\vee` for `||`, and `\neg` for `!`
2. Changed `\lt` and `\gt` to `<` and `>` (these commands were not defined using standard `amsmath`)
3. Skip applying `\mathrm` to function operator names when they're single characters (e.g., `f(x)` should be math-italicized and `loss(x)` should not)

By the way, great work with this package! It was super easy to extend it for my cases 😄

### DIFF
--- a/src/latexoperation.jl
+++ b/src/latexoperation.jl
@@ -158,11 +158,11 @@ function latexoperation(ex::Expr, prevOp::AbstractArray; cdot=true, kwargs...)
         if length(args) == 3
             # Check if already parsed elseif as args[3]
             haselseif::Bool = occursin(Regex("\\$textif"), args[3])
-            otherwise::String = haselseif ? "" : "& \\text{otherwise}"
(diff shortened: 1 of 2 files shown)


korsbo:
This looks super cool!

I really do not have time to go through everything but since it seems like a great improvement and it passes the tests I'll just risk it and merge without closer inspection. 

Thanks for this contribution!

korsbo:
Registering a version 0.14.0 now. It should be done in an hour or so.

PR #115, opened 2020-09-13
Superscript \prime and \rm math italics for words

mossr:
In unicode, the prime character  `′` is already positioned as a superscript, while in LaTeX it needs to explicitly be `^\prime` (so I've changed the `\prime`, `\backprime`, and other `\pprime` versions to be `raw"^\prime"`).

I also removed math italics for variables (i.e. words) longer than 2 characters (think `dx/dy` which should still be italicized).

Here is what it looked like *before* both modifications (notice the "lowered" prime and italicized words):
image

And here is what it looks like *after* both modifications:
image
 
Unrelated, but shout-out to [Pluto.jl](https://github.com/fonsp/Pluto.jl) which makes working with Latexify super satisfying!


korsbo:
This looks like an improvement. Some day soon it would be good to allow the user to decide what to italicize (with reasonable defaults). But not today. 

Thanks!

PR #121, opened 2020-11-03
Fix treatment of unicode sub- and superscripts

korsbo:


korsbo:
Replaces and closes #96

PR #129, opened 2020-12-01
Add Unitful support

gustaphe:
Add support for Untiful FreeUnits and Quantity, both in plain `\mathrm` mode and a more fancy `siunitx` style.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -28,6 +28,7 @@ Missings = "e1d29d7a-bbdc-5cf2-9ac0-f12de2c33e28"
 ParameterizedFunctions = "65888b18-ceab-5e60-b2b9-181511a3b968"
 SymEngine = "123dc426-2d89-5057-bbad-38513e3affd8"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [targets]
-test = ["Test", "DataFrames", "Missings"]
+test = ["Test", "DataFrames", "Missings", "Unitful"]
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -17,7 +17,8 @@ makedocs(
         "tutorials/recipes.md",
         "Use with other packages" => [
             "tutorials/parameterizedfunctions.md",
-            "tutorials/DiffEqBiological.md"
+            "tutorials/DiffEqBiological.md",
+            "tutorials/Unitful.md",
         ],
         "arguments.md",
         "tutorials/inner_workings.md",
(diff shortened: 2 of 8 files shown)


gustaphe:
It may be that this should have been recipes, I'm not familiar enough with those to make that work properly. Let me know if you think it would be worthwhile to attempt a rewrite.

korsbo:
Hi, I'm sorry it took so long for me to respond - the PR got lost in a deluge of other notifications. 

I like the idea of adding some nice support for this! I would have preferred to do this inside of Unitful but I can imagine that there are parts of creating that support that's not very intuitive. I'll have a look at your added tests to see what would be needed for this.

gustaphe:
No worries.
I actually had a bit of a look at it, because I realized my understanding of this package was a bit limited (like seriously, it's a bit of a monster, in a good way). I'll make an update to this PR for completeness, but I guess the proper place for it would be a PR to Unitful.jl.

gustaphe:
Turns out Unitful.jl are not very keen on keeping this functionality. They don't have a lot of dependencies, so adding Latexify or Requires would be a decent hit. So ehm, can it live with you? The alternative is creating a UnitfulLatexify package, but I'm not sure glue packages are the way to go, especially if we ever get that first class conditional support.

And if so, is the `julia_version<=1.0` errors something I should try to do something about?

korsbo:
Hi, sorry for ghosting you! My open-source work is really stretched thin right now. 

I too am a bit reluctant to add dependencies here. A while back, I realised that I'd likely need to support many packages in the future and that this would be unsustainable to do here. So, I too created a recipe system like that of Plots.jl (or, rather, copy-pasted a bunch from RecipesBase.jl). The idea was that having support for individual packages within Latexify.jl would eventually turn this into a behemoth of a package that's hard to maintain. I have already started making the other `@requires` imports obsolete by creating latexify recipes in the corresponding package instead. 

I'm still learning a bit as I go but so far the transition to remove support for other packages from Latexify.jl itself has made a lot of sense. Without doing so, I would for example be pretty much unable to simultaneously support two different breaking versions of the same package. Also, placing the recipes close to the code that eventually is being latexified makes it more natural for the devs who update the package to also update the latexify recipe. 

So far, I have still maintained latexify support for ModelingToolkit.jl, Catalyst.jl and ParameterizedFunctions.jl myself and since I essentially control both the latex recipes there as well as Latexify.jl, there have not been any issues with updating for breaking changes. Your request sort-of prompts the next phase of this - I get that Unitful.jl does not want to have to update a bunch of stuff whenever Latexify has a breaking change. I'm not sure what the optimal solution is but one thing that could make sense is to have a separate package for the latexify recipe framework - like RecipesBase.jl is. That would at least allow for a slow and conservative release schedule for the recipes framework while still having a quick-ish release schedule of Latexify.jl. 

I get that the maintainers of Unitful.jl is thinking just like me here - they too want to be a minimal dependency package. How did they solve the Plots.jl issue?

gustaphe:
Unitful uses an external package for plot recipes, so I decided to go that way too.

New package [UnitfulLatexify](https://github.com/gustaphe/UnitfulLatexify.jl) is on countdown for merge into Registry.
Thank you for your help on this.

PR #133, opened 2021-01-07
Support rendering DVI, PNG and SVG files from LaTeXStrings

### LINKED ISSUE #132: Support rendering PNG files
Hi @korsbo,

The [`render` function](https://github.com/korsbo/Latexify.jl/blob/005020d96d11cb57b4fcc97549eed69dfe479c11/src/utils.jl#L15) is very useful, but I wonder if it could be improved by supporting PNG output as well. `matplotlib.texmanager` does that by using `dvipng` (see code [here](https://github.com/matplotlib/matplotlib/blob/99e6240d95e86b4fd9edbcbefa8a5a135e707f27/lib/matplotlib/texmanager.py#L271) and [here](https://github.com/matplotlib/matplotlib/blob/99e6240d95e86b4fd9edbcbefa8a5a135e707f27/lib/matplotlib/texmanager.py#L332)).

Do you think it is a useful addition? I could try to make a PR.

schneiderfelipe:
This PR adds support for rendering DVI, PNG and SVG files from LaTeXStrings. This should close #132.

The format is specified by a MIME-type string and defaults to PDF (so nothing breaks):

```julia
latexify(:(x/y)) |> s -> render(s, MIME("image/png"))
```

jl_ZuyuWl

Supported MIME-type strings are:
- "application/pdf" (current behavior, requires LuaLaTeX)
- "application/x-dvi" (also requires LuaLaTeX - actually `dvilualatex`)
- "image/png" (requires [dvipng](http://www.nongnu.org/dvipng/))
- "image/svg+xml" (requires [dvisvgm](https://dvisvgm.de/))

I also wrote some docs, updated the doctring and added a small test for the generated TeX code. (This is my first contribution to a Julia package, yay!)


ChrisRackauckas:
Wow, that's awesome!

schneiderfelipe:
> [Coverage Status](https://coveralls.io/builds/36162040)
> 
> Coverage decreased (-1.6%) to 58.717% when pulling **[c21c7a7](https://github.com/korsbo/Latexify.jl/commit/c21c7a7ad9031d86e0a7a4c37fd8712dc592d4b2) on schneiderfelipe:latex-png** into **[6d1cea6](https://github.com/korsbo/Latexify.jl/commit/6d1cea67c74c3d59f100f259924eef1162428ac3) on korsbo:master**.

Coverage went down a bit. I'll add a simple test for checking the SVG output, since it's only text and should be easier to compare than DVI/PDF.

schneiderfelipe:
I just realized that there were no tests for the outputs, which makes perfect sense for PDF. I could either remove the SVG test (90045b2) or add LuaLaTeX and dvisvgm to be installed during CI. @korsbo, what do you think?

korsbo:
This looks Great! 

I'm just going through the PR now (sorry for the wait, I've been drowning in notifications). I think I'll have plenty of comments and thoughts, but over-all this looks really nice 👍

korsbo:
Before I go through details I'm thinking about some bigger picture things.
1. could/should we facilitate some degree of automatic rendering and showing of LaTeXStrings in for example VSCode? SVG can, for example, be displayed in the VSCode plot pane. I'm guessing that we could sometimes `Base.display` some of the generated images rather than open them. 
2. backwards compatibility is not really required here - if we find a solution that is much better than before then I'm perfectly willing to make a breaking release. 
3. rendering LaTeXStrings is probably useful for more than just Latexify. Should this be broken out into its own package? 

Just some initial thoughts - I'll keep on playing around.

src/utils.jl
```diff
@@ -35,20 +30,97 @@ function render(s::LaTeXString; debug=false, name=tempname(), command="\\Large")
...
+    render(s, MIME("application/x-dvi"); debug=debug, name=name, command=command, open=false)
+
+    cmd = `dvisvgm -n -v 0 -o $(name).svg $(name).dvi`
+    debug || (cmd = pipeline(cmd, devnull))
+    run(cmd)
+
+    if open
+        _openfile(name; ext="svg")
+    end
+
```
korsbo, line 123: Adding 
```julia
    if callshow && displayable(MIME("image/svg+xml"))
        Base.open(fname * ".svg") do f
            display(MIME("image/svg+xml"), read(f, String))
        end
    end
```

and a corresponding `callshow` kwarg to the function would allow for displaying the SVG in the user's environment (for example the plot pane in vscode). 

I know that this would work nicely with VSCode and SVG, but I'm guessing that it'll work for some other mime-types as well.  

This can probably be done with a nicer interface though. I'm thinking that it could be nice to display the file in the user's environment by default (if possible) and to fallback to opening the file externally.

korsbo:
I should add that the PR works well as it is and that I would not mind just merging it as is - leaving improvements for future PRs - if that would suit you better!

korsbo:
> I just realized that there were no tests for the outputs, which makes perfect sense for PDF. I could either remove the SVG test ([90045b2](https://github.com/korsbo/Latexify.jl/commit/90045b28cd1480656b1f0be44fcc57d200bf92c5)) or add LuaLaTeX and dvisvgm to be installed during CI. @korsbo, what do you think?

If you could add the dependency installation to the CI and leave the test, that'd be great!

schneiderfelipe:
> Before I go through details I'm thinking about some bigger picture things.
> 
> 1. could/should we facilitate some degree of automatic rendering and showing of LaTeXStrings in for example VSCode? SVG can, for example, be displayed in the VSCode plot pane. I'm guessing that we could sometimes `Base.display` some of the generated images rather than open them.

That's a good idea. I'll try to make it work in Pluto.jl.

> 2. backwards compatibility is not really required here - if we find a solution that is much better than before then I'm perfectly willing to make a breaking release.

Right. One breaking thing would be to default to `Base.display` instead of opening them. A keyword could control whether to open the file or not. Is that good?

> 3. rendering LaTeXStrings is probably useful for more than just Latexify. Should this be broken out into its own package?

That's an interesting idea. I think this fits Latexify.jl, but someone could consider it a stretch to have Latexify.jl as a dependency just for the rendering.

> Just some initial thoughts - I'll keep on playing around.

src/utils.jl
```diff
@@ -35,20 +30,97 @@ function render(s::LaTeXString; debug=false, name=tempname(), command="\\Large")
...
+end
+
+
+"""
+    render(::LaTeXString[, ::MIME"mime"]; debug=false, name=tempname(), command="\\Large", open=true)
+
+Display a standalone document with the given input. Supported MIME-type strings are
+"application/pdf" (default), "application/x-dvi", "image/png" and "image/svg+xml".
+"""
+function render(s::LaTeXString; debug=false, name=tempname(), command="\\Large", open=true)
```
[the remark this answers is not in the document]
schneiderfelipe: Great!

src/utils.jl
schneiderfelipe: That's nice to defer to `displayable` and support all mime-types this way. I'll default to `display`ing if possible and opening if not, in that order of priority, but leaving `callshow=true, open=true` as keywords if the user wants more control.

schneiderfelipe:
I think I've added (almost) everything.

It now works in Pluto.jl, for instance,

Screenshot from 2021-01-18 02-22-14

Observe that `render` is not directly called. This works by defining a `Base.show(io::IO, ::MIME"image/png", s::LaTeXString)` that actually calls `render`. This `Base.show` is then called by the display system.

(Note: I attempted to write an equivalent `Base.show` for SVG, but I couldn't make it work - not in Pluto.jl at least -, it shows

```
Failed to show value:

MethodError: no method matching show(::Base.PipeEndpoint, ::MIME{Symbol("image/svg+xml")}, ::String)
```

Maybe @fonsp could help. I couldn't figure out how to make a `String` be interpreted as `"image/svg+xml"`.)

In the REPL, `render` still opens files:

```julia
julia> using Latexify

julia> latexify(:(x/y)) |> s -> render(s, MIME("application/pdf"))

julia> latexify(:(x/y)) |> s -> render(s, MIME("image/png"))

julia> latexify(:(x/y))
L"$\frac{x}{y}$"
```

schneiderfelipe:
I like using the display system and not requiring `callshow`, etc. But there is actually another (minor) issue with SVG in the REPL if we pursue using `if callshow ... display(...)` though: it was a surprise to me to discover that the REPL can "display" SVG (well, not really):

```julia
julia> displayable(MIME("application/pdf"))
false

julia> displayable(MIME("application/x-dvi"))
false

julia> displayable(MIME("image/png"))
false

julia> displayable(MIME("image/svg+xml"))  # !
true
```

This has to do with the fact that it is a plain text format (the same thing happens with `"text/html"`, [see here](https://discourse.julialang.org/t/possibly-incorrect-displayable-text-html-in-v0-6/2145/3?u=schneiderfelipe)). As such, `displayable` wouldn't be reliable in the REPL, at least for SVG output.

korsbo:
This looks cool! But I think that you have introduced a type-piracy issue when overloading `Base.show` since Latexify does not "own" `LaTeXString`. This means that `LaTeXStrings` will work differently depending on whether `Latexify` has been imported. Worse yet, if the user does not have the right deps installed, loading Latexify.jl could break LaTeXStrings.jl. If we want full control over how the return type of `latexify` is dealt with then we'd need to make our own type - but the trade-off is that this would likely work less well with other packages. I realise that I was a bit unclear about this in the suggestion that I made, sorry!

My idea was to still require the user to call `render` and not to rely on a `show` overload. However, `render` can be clever and default to the best possible way of displaying the output to the user. Thus, if the current environment allows for example `MIME"image/svg+xml"` then `render` could call `display(MIME("image/svg+xml), svg_str)"`. If the environment does not allow for displaying images, then `render` could call for an external viewer. And, if the user wants to override the default then we could have some kwarg for that. 

-------

> This has to do with the fact that it is a plain text format (the same thing happens with "text/html", see here). As such, displayable wouldn't be reliable in the REPL, at least for SVG output.

That'll make automatic inference of display mode a bit more tricky... I wonder if there's some way of figuring out whether the svg will be displayed as an image or as text. I'm guessing so,  but I have not (yet) looked into it.

--------

I don't really use Pluto.jl (I'd miss my vim keybindings too much) but does not Pluto have Mathjax support? Does not rendering work automatically for LaTeXStrings? Is there then still some use-case for displaying the equations as an image or might we just leave it?

schneiderfelipe:
> This looks cool! But I think that you have introduced a type-piracy issue when overloading `Base.show` since Latexify does not "own" `LaTeXString`. This means that `LaTeXStrings` will work differently depending on whether `Latexify` has been imported. Worse yet, if the user does not have the right deps installed, loading Latexify.jl could break LaTeXStrings.jl. If we want full control over how the return type of `latexify` is dealt with then we'd need to make our own type - but the trade-off is that this would likely work less well with other packages. I realise that I was a bit unclear about this in the suggestion that I made, sorry!

Right, then it's a bad idea. Just some thought, what if a new package were created (`LaTeXStringRenderer.jl` or something)? I suspect that would still be type-piracy?

> My idea was to still require the user to call `render` and not to rely on a `show` overload. However, `render` can be clever and default to the best possible way of displaying the output to the user. Thus, if the current environment allows for example `MIME"image/svg+xml"` then `render` could call `display(MIME("image/svg+xml), svg_str)"`. If the environment does not allow for displaying images, then `render` could call for an external viewer. And, if the user wants to override the default then we could have some kwarg for that.

I actually did that at first, but then things wouldn't render automatically. Now I see how your concerns match the issue above. 🙈

> I don't really use Pluto.jl (I'd miss my vim keybindings too much) but does not Pluto have Mathjax support? Does not rendering work automatically for LaTeXStrings? Is there then still some use-case for displaying the equations as an image or might we just leave it?

Yes, MathJax is supported when written inside Markdown, but LaTeXStrings are not rendered (but you could concatenate them as strings inside Markdown).

Anyway, I'll revert my recent changes and call `display` as you initially suggested.

schneiderfelipe:
OK, I reverted things so that there's no overload of `Base.show` and things like

```julia
julia> using Latexify

julia> latexify(:(1 / (σ * sqrt(2π)) * exp(-((x - μ) / 2σ)^2))) |> render

julia> latexify(:(1 / (σ * sqrt(2π)) * exp(-((x - μ) / 2σ)^2))) |> s -> render(s, MIME("image/png"))
```

open files in the REPL. But I removed the `callshow` feature for SVG since I couldn't make it work properly in the REPL (because of [this comment](https://github.com/korsbo/Latexify.jl/pull/133#issuecomment-761998242)). Do you have a suggestion for this?

PNGs are shown in VSCode in the Plots pane:

Screenshot from 2021-01-20 23-27-22

---

By the way, I was wrong, Pluto.jl does support rendering LaTeXStrings using MathJax:

Screenshot from 2021-01-20 22-41-09

schneiderfelipe:
> If you could add the dependency installation to the CI and leave the test, that'd be great!

I added the test, but the one that checks SVG output only runs on Linux (I realized installing LaTeX and such could be difficult in other platforms). Please tell me if that's too bad.

korsbo:
This looks awesome! 

I asked about the `displayable` issue on the Julia slack and got an answer! `displayable(MIME("image/svg+xml))"` returns true even if it is only displayable as text. `displayable(MIME("image/svg"))` only returns true if displayable as an image. Currently this is broken in VSCode as `displayable(MIME("image/svg"))` erroneously returns false. However, that is already fixed in master and it should all work upon their next release. We can thus safely use this check and it'll start working in vscode when we get the new version.

> I added the test, but the one that checks SVG output only runs on Linux (I realized installing LaTeX and such could be difficult in other platforms. I have no clue how to do it on Windows 🙈). Please tell me if that's too bad.

That's fine! I might consider adding some tests using https://github.com/JuliaPlots/VisualRegressionTests.jl in the future but I have yet to figure out how all that works so we don't have to do that right away.

schneiderfelipe:
Nice! Now all lines below work in VSCode as expected:

```julia
using Latexify

tex = latexify(:(1 / (σ * sqrt(2π)) * exp(-((x - μ) / 2σ)^2)))

tex |> render  # opens PDF
tex |> s -> render(s, MIME("application/pdf"))   # opens PDF
tex |> s -> render(s, MIME("application/x-dvi"))   # opens DVI
tex |> s -> render(s, MIME("image/png"))   # shows PNG in Julia Plots pane
tex |> s -> render(s, MIME("image/svg"))  # shows SVG in Julia Plots pane
```

In the REPL, all of them open files 👍🏽 

It turns out that we want to check for the displayability of `MIME("image/svg")`, but we want to actually display `MIME("image/svg+xml")`, otherwise it won't work in VSCode.

korsbo:
> It turns out that we want to check for the displayability of MIME("image/svg"), but we want to actually display MIME("image/svg+xml"), otherwise it won't work in VSCode.

I think that this will be fixed in the next release. We should then be able to use `display(MIME("image/svg"), svg_str)`. AFAIS, it should not make any difference in terms of functionality, it will just serve to make the code less confusing. I can change this afterwards, when the vscode extension update is actually released.

korsbo:
So, are we done, or do we want to add some automated default selection of display type? Like have a priority list where if the first mime type is supported use that one, if not try the next, and so on until you get to a file which opens externally.

korsbo:
Just so you know, I'm perfectly happy with calling this done and leave changes for the future if you'd prefer to drop this for now. It's nice to collapse multiple breaking changes into one, but so far the changes have not been breaking so there's no problem with releasing it bit-by-bit.

korsbo:
🎉 Done! 🎉 

I'll just check this out locally to play around a bit, merge https://github.com/korsbo/Latexify.jl/pull/134, and then I'll make a new release! 

Thanks for this improvement!

PR #135, opened 2021-02-02
Recursive array latexification

gustaphe:
Uses `latexify(..., env=:raw)` rather than `latexraw(...)` to invoke latexrecipes on the elements of arrays. 

(I need this for UnitfulLatexify)

### DIFF
--- a/src/latexarray.jl
+++ b/src/latexarray.jl
@@ -23,11 +23,7 @@ function latexarray(arr::AbstractArray; adjustment::Symbol=:c, transpose=false, 
     str = "\\left[\n"
     str *= "\\begin{array}{" * "$(adjustment)"^columns * "}\n"
 
-    arr = latexraw(arr; kwargs...)
-    for i=1:rows, j=1:columns
-        str *= arr[i,j]
-        j==columns ? (str *= eol) : (str *= " & ")
-    end
+    str *= join(join.(eachrow(latexify.(arr;kwargs...,env=:raw))," & ") .* eol)
 
     str *= "\\end{array}\n"
     str *= "\\right]"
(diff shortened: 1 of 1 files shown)


korsbo:
It seems that you once again have fallen victim of my notification overload - sorry I did not see this before now.

I'm curious, why do you need that? Is it due to the latexrecipes acting only on the latexify function? You needing this might indicate that I have a design-flaw.

korsbo:
I'm in the slow-ish process of refactoring latexify and making it a bit more sane for extensions so I'm hoping that this should not be needed once I'm done. 

Until then though, we might as well apply this quick-fix.

korsbo:
It seems that the particular way you have written this however is breaking latexify for old julia versions where `eachrow` was not a thing. Could you change it to something like

```
    for i=1:rows, j=1:columns
        str *= latexify(arr[i,j]; env=:raw, kwargs...)
        j==columns ? (str *= eol) : (str *= " & ")
    end
```
so that we can keep supporting ancient Julia versions?

gustaphe:
There's actually a couple of these things where the recursion isn't quite working (`Expr`s, for instance), but I'm not quite confident enough in the inner workings to tell if all uses of `latexraw` should be replaced with `latexify(...;env=:raw)`, or if `latexraw` should be updated to try latex recipes (I'm afraid of causing infinite loops of `latexify(latexify(...(::Int64)))`). Or whatever the corresponding idea would be in the refactored version.

But I don't want to add any work, especially if a new system is in the works, it's not like UnitfulLatexify is being used at any large scale. LMK if I can help at all.

Ah, backwards compatibility, that everlasting curse. Sorry, I've reverted that line.

korsbo:
I'm working with a prototype that allows far greater control over how parts of an expression are treated (you should be able to find the branch here). Hopefully it'll all become more sensible then.

But until then, this'll do. Cheers!

PR #138, opened 2021-03-04
Route exported functions through latexify

gustaphe:
(I know you're working on a big rework, so depending on the time scale it might not be worthwhile to merge this but...)

As you know, I've wanted to sort out the recursion in this package for a while. Essentially, `latexify()` runs its arguments past the recipe system, but the subcommands like `latexraw` don't. I've previously tried replacing these by `latexify(..., env=:raw)` etc, but it turned out to be a lot of edits, and I think it makes sense to export and document the likes of `latexraw` for user use.

But if you make `latexraw(...)` a synonym for `latexify(..., env=:raw)`, you'll get stuck in a loop. *Unless*...

With this pull request, all of the exported functions are synonyms for `latexify()` with the corresponding `env` set. Internally, `latexify` uses the unexported functions `_latexraw` and the like.

For users this makes no difference (hence no docs change). It *did* break one of my recipes, but one that was only working because the functionality I want wasn't there (It called `latexraw(x::MyType)` inside the recipe for `MyType`).

I also commented out the svg test, because it kept failing because of version issues, and it doesn't seem like a very well-formed test.

### DIFF
--- a/src/Latexify.jl
+++ b/src/Latexify.jl
@@ -29,7 +29,6 @@ include("unicode2latex.jl")
 include("function2latex.jl")
 include("latexraw.jl")
 include("latexoperation.jl")
-include("latexify_function.jl")
 include("latexarray.jl")
 include("latexalign.jl")
 include("latexbracket.jl")
@@ -40,14 +39,15 @@ include("default_kwargs.jl")
 include("recipes.jl")
 include("macros.jl")
 
-include("md.jl")
 include("mdtable.jl")
 include("mdtext.jl")
+include("md.jl")
 
 include("utils.jl")
 
 include("numberformatters.jl")
 
+include("latexify_function.jl")
 
 ### Add support for additional packages without adding them as dependencies.
 function __init__()
--- a/src/latexalign.jl
+++ b/src/latexalign.jl
@@ -39,9 +39,9 @@ julia> latexalign(ode)
 ```
 
 """
-function latexalign end
+latexalign(args...; kwargs...) = latexify(args...; kwargs..., env=:align)
 
-function latexalign(arr::AbstractMatrix; separator=" =& ", double_linebreak=false, starred=false, rows=:all, aligned=false, kwargs...)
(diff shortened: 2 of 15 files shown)


korsbo:
This seems fine but I'm not entirely sure just what problem it solves. Does it affect how recursion is done at all or does it only affect what happens when a user calls `latexraw`, etc?




> I also commented out the svg test, because it kept failing because of version issues, and it doesn't seem like a very well-formed test.

that's fine. This one has been bugging me as well.

gustaphe:
An example from UnitfulLatexify:

```julia
using Unitful,Latexify,UnitfulLatexify
x = 23u"m"
latexify(:(x=$x))
```
currently errors, because `latexify(::Expr)` calls `latexinline(::Expr)` which calls `latexraw()` for each of the components. And `latexraw()` isn't defined for `x::Unitful.Quantity`.

With these changes, `latexify(::Expr)` calls `_latexinline(::Expr)` which calls `latexraw()` for each of the components. `latexraw()` is now a synonym for `latexify(...,env=:raw)`, which is of course defined and applies any available recipes. In the case of `x::Unitful.Quantity`, the recipe means that `latexify` finally calls `_latexraw` for a defined type, and the final result is 

```julia
L"$x = 23\;\mathrm{m}$"
```

Yes, it affects how the recursion is done, in that it loops many of these calls back to the top level `latexify` command. So it changes what happens when a user calls `latexraw`, but also what happens when `latexify` internally calls `latexraw`. `_latexraw` is now the low-level definition of how the basic types are latexified. 

The other way to solve this problem would be to replace every call to `latexraw` etc in the package by literal `latexify(...,env=:raw)` etc., and document to users why they should use `latexify(...,env=:raw)` rather than `latexraw()` if the argument could have a defined recipe. I think this pr is cleaner.

korsbo:
Ok, I see now, thanks! 

This looks like a nice update to me - merging!

korsbo:
huh, my conflict resolution seems to have landed in an odd place. Should be fine but I'll wait for the CI to finish before merging.

korsbo:
Released as 0.14.8 https://github.com/JuliaRegistries/General/pull/31342

PR #139, opened 2021-03-05
Surround expr tuple with parentheses.

korsbo:


korsbo:
My tests are broken locally (#140) so I'll use the CI, for now, to figure out whether this is causing regressions.

korsbo:
Urgh, that messes up function calls. Dropping this for now.

Issue #144, opened 2021-03-05
latexifymerge


gustaphe:
https://github.com/korsbo/Latexify.jl/blob/36f88564dffe40005fefa68cdaae9b0135a73812/src/latexoperation.jl#L16-L18

Is this intended for recipes to be able to return a series of arguments whose latexifications are just concatenated? If so, that's very useful functionality, but then there's a bug. If not, I would like to request such functionality.

It seems the other uses of `latexoperation()` return `String`s, but this one returns a `Symbol`, so 
```julia
julia> latexify(Expr(:latexifymerge,3,"cm"))
ERROR: MethodError: no method matching unicode2latex(::Symbol)
```

Simply removing the `Symbol()` at L18 seems to fix this, but I'm not completely confident this is what that function is supposed to do.

korsbo:
You're right. It is intended to allow for easy joining of things. Is should probably be `join(args)` though.

PR #146, opened 2021-03-07
Shift equation<-array<-raw for arrays

gustaphe:
Right now, `latexarray(::Array)` puts an `equation` around an array, and there is a pattern that `latexraw(x::Arr) == [ latexraw(x[1]), latexraw(x[2]), ... ]` for some of the commands. I think it would make more sense to have `latexarray` *just* do the array, `latexraw(::Array)` call `_latexarray`, and use broadcasting (`latexraw.(arr)`) to latexify each item of `arr`. This makes it easier to dispatch on `Array` types in recipes. 

`latexify(::Array)` now infers that you want to use `_latexequation`, rather than `(...) -> _latexequation(_latexarray(...))`. Since `latexraw(::Array)` now does the brackets and everything, when `_latexequation()` calls `latexraw` it will successfully latexify the array.

Small changes in usage, see how the tests have changed.

Also included: The `Symbol` change from #144

### DIFF
--- a/docs/src/tutorials/rendering_latex.md
+++ b/docs/src/tutorials/rendering_latex.md
@@ -9,4 +9,4 @@ display("text/latex", x)
 ```
 where `x` is a latex-formatted string.
 
-This requires `x` to specify a ``\LaTeX`` environment. `latexalign` and `latexarray` already does this, but if you want to render the result of `latexify` you must supply an environment (for example `"\$ $x \$"`).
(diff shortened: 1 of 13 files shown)


korsbo:
I think that I like this! It is however breaking and I'd need to spend some time getting the document deployment working properly again before I can merge this.

korsbo:
This change makes plenty of sense! Merging now.

Thank you for your contribution(s)! 🎉

PR #156, opened 2021-03-31
Replace Appveyor with GH actions.

korsbo:


korsbo:
And, ironically, this PR is triggering Appveyor but not GH. Sigh...

korsbo:
There we go.

korsbo:
Urgh, the [newline parsing bug](https://github.com/JuliaLang/julia/issues/38908) is doing its thing here.

korsbo:
Fixed in #157 instead

PR #157, opened 2021-04-01
Internal performance and CI tweaks

### LINKED ISSUE #140: Windows testing bug
There's a bug in Julia that affects how newlines are parsed (https://github.com/JuliaLang/julia/issues/38908) and that one is causing Latexify tests to fail on Windows. 

The best fix would be in Julia itself but that could take a while. In the meantime, we could use `replace` to remove any strange newline characters that might spuriously pop up in windows.

korsbo:
Hopefully improves compile time a little.

- fixes tests for windows (closes #140)
- switches to GH actions for windows CI.


korsbo:
I converted the original scope of this PR so that I could merge into master and use all these useful things in more experimental compile-time fixing branches.

PR #158, opened 2021-04-01
Don't double broadcast

gustaphe:
This reverts a change I made in #146. Kind of obscure difference, it seems most of the time `i` here is either a string or some kind of scalar anyway. I can't even formulate a test for this without pulling in a dependency for Unitful. But it seems to me that the intention is not to broadcast into `i`, and this doesn't break any tests so...

(If you're curious, for some reason there are types in Unitful that are broadcastable but don't have a `length` method, so `latexraw.(i)` errors out for them. I'm opening an issue for this over there too, but irrespective I think this is the intended behaviour here)

### DIFF
--- a/src/latexoperation.jl
+++ b/src/latexoperation.jl
@@ -6,9 +6,9 @@ This uses the information about the previous operations to decide if
 a parenthesis is needed.
 
 """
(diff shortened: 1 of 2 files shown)


gustaphe:
No, I could think of a simple test that crashes without this fix and passes with it.

PR #163, opened 2021-04-12
Add refactor prototype overloading `show`.

korsbo:
This is a competing refactorisation to #161 

It overloads `Base.show` with `MIME"text/latexify"`. This could allow others to extend latexify's functionality without ever even loading Latexify.jl itself. 

It's based on suggestions by @gustaphe

### DIFF
--- a/refactor_show_prototype.jl
+++ b/refactor_show_prototype.jl
@@ -1,0 +1,136 @@
+using Pkg
+Pkg.activate(@__DIR__)
+using Latexify
+import Base: show
+using SnoopCompile
+
+
+const Lmime = MIME"text/latexify"
+const lmime = MIME("text/latexify")
+
+const opdict = (;
+    :* => :Multiplication,
+    :/ => :Division,
+    :+ => :Addition,
+    :- => :Subtraction,
+    :^ => :Pow,
+)
+for T in values(opdict)
+    println(T)
+end
+abstract type AbstractLatexifyOperation end
+for T in values(opdict)
+    @eval(
+        struct $(T){S} <: AbstractLatexifyOperation
+            args::S
+        end
+    )
+end
+
+const ops = (; (keys(opdict) .=> eval.(values(opdict)))...)
+ops[:*]([:a, :b])
+
+for op in pairs(ops)
+    println(op[1], op[2])
+    @eval latexop(::Val{:call}, ::Val{$(QuoteNode(op[1]))}, args) = $(op[2])(args)
+end
+
+
+abstract type AbstractLatexifyEnvironment end
+for T in [:Align, :Equation, :Inline, :Bracket, :Raw, :Table]
+    @eval(
+        struct $(T){S} <: AbstractLatexifyEnvironment
+            args::S
+        end
+    )
+end
+Align(x, y) = Align(hcat(x, y))
+
+
+
+latexop(ex::Expr) = latexop(Val(ex.head), Val(ex.args[1]), ex.args[2:end])
+latexop(x) = x
+latexop(x...) = x
+latexop(x::Matrix) = Align(x)
+latexop(::Val{:align}, x) = Align(x)
+latexop(::Val{:equation}, x) = Equation(x)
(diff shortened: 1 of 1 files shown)


korsbo:
One potential problem with this approach is that Latexify.jl does not own `Base.show` and can thus not precompile it.

refactor_show_prototype.jl
```diff
@@ -0,0 +1,136 @@
...
+latexop(x) = x
+latexop(x...) = x
+latexop(x::Matrix) = Align(x)
+latexop(::Val{:align}, x) = Align(x)
+latexop(::Val{:equation}, x) = Equation(x)
+latexop(lhs::Vector, rhs::Vector) = Align(lhs, rhs)
+
+show(io::IO, ::MIME"text/latexify", x) = print(io, x)
+
+compare_precedence(a, b) = Base.operator_precedence(a) > Base.operator_precedence(b)
```
gustaphe, line 61: I'm not entirely sure this will do the right thing for things like `1 - (2 + 3 + 4)` -- here the precedence is equal, but `compare_precedences` will return `false`. Perhaps that is solved by changing the `>` to `>=`, but will it then do the right thing for `1 - (2 - 3 - 4)`?

I guess the `show` method for `:-` could give `:+` as precedence for left and `:-` for right.

korsbo: I agree, this will not always do the right thing

korsbo: I'm not even convinced that `operator_precedence` is reliably good for latex purposes. I think one might find plenty of cases where code would need parentheses while latex would not.

refactor_show_prototype.jl
```diff
@@ -0,0 +1,136 @@
...
+    println(T)
+end
+abstract type AbstractLatexifyOperation end
+for T in values(opdict)
+    @eval(
+        struct $(T){S} <: AbstractLatexifyOperation
+            args::S
+        end
+    )
+end
```
gustaphe, line 28: An alternative is 
```julia
struct Operation{S, T} where {S<:Val, T}
args::T
end
```
and the likes of
```julia
function show(io::IO, mime::Lmime, Operation{Val{:+}, T}) where T
```

refactor_show_prototype.jl
gustaphe: If there exists such an ordering, it's pretty simple to implement your own `operator_precedence`. Otherwise one might need individual `compare_precedence(::Val, ::Val)` functions.

gustaphe:
> One potential problem with this approach is that Latexify.jl does not own `Base.show` and can thus not precompile it.

I haven't worked with precompilation, but don't you just need to own the methods? `precompile(show, (IO, MIME"text/latex", Int64))` should be uncontroversial. Or am I misunderstanding?

refactor_show_prototype.jl
korsbo: individual comparisons is pretty much what I have been doing before (not saying that this it the way to go). It turned out not to be needed too frequently though. `-` and `^` turned out to be the ones which needed the most care in figureing out where parentheses are needed.

refactor_show_prototype.jl
korsbo: Good idea. I modified it a bit though. I added the ability to overload on `head` and I removed type specification on the args since I don't want the compiler to over-specialise on that (faster runtime at the expense of compiletime).

korsbo:
One advantage of #161 over this approach is that it would be possible to let the list of matching functions be passed in as a `kwarg` or something that is specific to a single call. 

Here, all extensions are global and the overloads created in one package might bleed over and affect others. If a user wanted to, for example, strip away all macro call annotations they could overload `show(...  x::Operation{Val{:macrocall}, T}) where T = show(io, mime, args(x)[end])`. That would work nicely but it would affect the latexification done in other packages too. Having users choose between piracy and simply not being able to do what they want is not ideal. 

In #161 it might be possible to let extending packages do what they want while keeping the effects confined to only apply when the top-level call takes their own types. There could then be two modes of extension - local or global (for things like UnitfulLatexify.jl). Here, only global is possible.

PR #167, opened 2021-04-14
Addresses issue with non-numerical sub- and super-scripts.

akabla:
This patch allows us to use a broader range of Unicode symbols with sub-scripts and super-scripts. I was having issues with Greek letters in sub-script in one of my projects and this is fixing the problem. See the example below:

```
julia> latexify(:(2a₍ₐᵦᵧ₎+ΓhΛ₁₂- ħᵗⁱᵦ))
L"$2 \cdot a{_{( a \beta \gamma )}} + {\Gamma}h\Lambda{_{1 2}} - \hbar{^{t i}}{_\beta}$"
```
Many thanks for this great library btw!


gustaphe:
This is very slightly wrong. It needs to be `a_{\gamma}` rather than `a{_\gamma}` etc. The former will adapt to the size and shape of the left hand argument, the latter will act as if you've written `a{}_{\gamma}`.

```tex
\[
\Big){^5}{_5}\quad\Big)^{5}_{5}
\]
```

image

akabla:
I see some tests are failing. It seems to be due to the fact that I inserted spaces between symbols to avoid problems with commands that are part of the sub/super script. This changes the tex output, but not the result once tex is compiled. Tests should be easy to fix accordingly if you like the changes made.

akabla:
> This is very slightly wrong. It needs to be `a_{\gamma}` rather than `a{_\gamma}` etc. The former will adapt to the size and shape of the left hand argument, the latter will act as if you've written `a{}_{\gamma}`.
> 
> ```
> \[
> \Big){^5}{_5}\quad\Big)^{5}_{5}
> \]
> ```
> 
> image

Interesting - that is a subtlety I was not aware of! Let me see if I can sort it out.

gustaphe:
The problem is that if you have several sub- or superscripts in a row you will get issues either way. The correct way to deal with that would be to take characters `while` they are  `in` [a predefined list of sub- or superscript characters](https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts#Superscripts_and_subscripts_block) and then put that entire list inside the gull wings for a single `_` or `^`.

But at least the characters that currently get converted to `\_x` can without controversy be converted to `_{x}`.

akabla:
@gustaphe 
I think this is now more compliant with what you said. Symbols are indeed all grouped under the same block with a single _ or ^.
```
julia> latexify(:(2a₍ₐᵦᵧ₎+ΓhΛ₁₂- ħᵗⁱᵦ))
L"$2 \cdot a_{( a \beta \gamma )} + {\Gamma}h\Lambda_{1 2} - \hbar^{t i}_{\beta}$"
```
Tests are now passing fine too. Note I had to change the bit the form of the expected LaTeX to be produced.

akabla:
There were a few edge cases that made the approach more complex than I thought. It now works well for what I needed and feels a lot more robust for the situations I tested (but I may have been overlooking other aspects). 
```
julia> latexify( :(dᵝdtᵝ(dᵃ₂)) )
L"$\mathrm{d^{\beta}dt^\beta}\left( d^a_2 \right)$"
```

There is however a test case that can't be resolved very well:

@test latexraw("β₃_hello") == raw"\beta{_3}_{hello}"

With the patch I propose, the string `"\beta_3_{hello}"` would be formed, which does not compile well. `\beta{_3}_{hello}` is also strange, as hello would be the subscript of 3 I think. I'd guess we would want it to be interpreted as `"\beta_{3 hello}"` but I'm not sure how to get there. I don't know how much of this test case is relevant. If this is a problem, it is easy to keep the curly brackets around the expression as in the first version I proposed. In any case, I'd very much welcome some support for the non-numerical superscript as part of the library.

gustaphe:
Technically, it should probably be `\beta_{3\textrm{hello}}`, but I don't think there's an efficient algorithm to figure out if a subscript is a collection of variables or a text string.

If, as a first intermediate step, `r"_[^\^_]+" => {\0}`, (or something like that) then that special case should be solved too, right? 

Another edge case, which shouldn't happen but could, is `\beta{_a}{^b}{_c}`. To solve that, one could do `=> \2\1` while it matches `{^...}{_...}` to get subscripts first and superscripts after. A more efficient way is probably to build a sub vector and a sup vector. That would probably also simplify the merging procedure. 

(Btw I'm not a maintainer, just someone who texes a lot)

korsbo:
Thanks for the PR @akabla and thanks for all the good advice and insights that you bring @gustaphe! 

It's funny how enabling latexification of almost arbitrary expressions turned out to be a lot simpler than just getting subscripts right. You're not the first one to step on this landmine though. 

I have a feature-freeze at work today(-ish) - so lots to do - but I might be able to have a look tomorrow or Monday. 

Note that I have not looked at this properly yet and do not take my words as gospel, you two seems to be interating out a bunch of kinks just fine on your own! A few high-level pointers though. 
- I'm rewriting Latexify.jl as we speak - some things will change and it is also not a bad time to introduce breaking changes if they seem necessary. I had not planned on changing `unicode2latex` but `convertSubscript` is more likely to need some love. 
- One problem here is the mixing of unicode subscripts and the treatment of underscores within symbols. I've never been very happy about how underscores are treated though. `X_1` should probably be `X_{1}` whereas `hello_world` should probably be `hello\_world`. Adding further underscores adds to the complexity `X_1_2` `my_long_function`. I don't know of a general solution that does not have trade-offs here. Nor do I know of a way to discriminate between the different cases in robust ways. If it turns out that the trade-offs to the current solution (lower the first underscore, escape the rest) is causing too much problems then that might be up for revision.

korsbo:
Note also that the execution order of `convertSubscript` and `unicode2latex` could be changed if this would help.

gustaphe:
Hm, I'm not sure when you would ever want an underscore in latexification (except in code listing, ``\texttt{hello\_world}``). If your variables are snake-cased, you rename them for latexification anyway imo.

gustaphe:
New problem: How about `latexify(:(a³ ^ 3))`? (I've never used the unicode sub- and superscript characters and I'm increasingly convinced they are evil)

Perhaps the initial idea of `{^3}` isn't as bad as I first thought. It'll be ugly in some cases, and it's wrong latex-wise, but it will avoid compile errors in cases like that...

korsbo:
> Hm, I'm not sure when you would ever want an underscore in latexification (except in code listing, `\texttt{hello\_world}`). If your variables are snake-cased, you rename them for latexification anyway imo.

I think Latexify.jl occupies a somewhat mixed niche. My original intent was to avoid having to transcribe the massive systems of differential equations that a student of mine had left behind. It was all about getting to the latex code for inclusion in a latex document. Now, my primary use is not really for generating documents but rather for visualising my code or DSL-created maths objects while coding. Here, I do find myself latexifying things that were written using programming notation rather than maths notation. This has turned out the be the greatest utility of latexify in my work. The different use of underscores in programming vs latex-maths is therefore a bit tricky.

akabla:
> Perhaps the initial idea of `{^3}` isn't as bad as I first thought. It'll be ugly in some cases, and it's wrong latex-wise, but it will avoid compile errors in cases like that...

I think that's right; maybe this is the simplest way forward indeed. I have to say that @gustaphe convinced me that `A{^{1 i}}{_{2 j}}` doesn't look as nice as `A^{1 i}_{j 4}` in terms of the output. But avoiding errors takes priority in my view. 

In the PR code, the `{}` surrounding the latex expressions are removed at the end of each merge function, so I think it's just a matter of getting rid of a few lines of code to get back to the `{^3}` output, if this is what is preferred.

Issue #174, opened 2021-05-27
latexify(-Inf) ignores the -


jessebett:
```
julia> using Latexify

julia> latexify(Inf)
L"$\infty$"

julia> latexify(-Inf)
L"$\infty$"
```

![Screen Shot 2021-05-26 at 10 27 36 PM](https://user-images.githubusercontent.com/4240549/119756785-fa02c700-be71-11eb-8733-887097453c79.png)

korsbo:
Fixed in v0.15.6

PR #175, opened 2021-05-27
catch -Inf

jessebett:
This addresses #174


korsbo:
Thanks for submitting a fix! I did not notice this PR before I submitted my own to address the same issue. 

Where you use 'x == -Inf', I use 'sign(x)' and 'isinf(x)'. I have been conditioned to avoid '==Inf' in favour of 'isinf' so I would imagine that the latter solution is "safer". I'll merge #176 and close this one but thanks nevertheless!

PR #177, opened 2021-06-07
Update latexoperation.jl to correctly process "//" in expressions as fractions

sergeyfarin:
1//2 processed as a rational number; however "1//2" in expressions processed as //(1,2). 
Effectively for latexify // in expressions should be equivalent to /.


korsbo:
That makes sense, thanks!

PR #184, opened 2021-10-07
link_from_docs_to_repo

ranocha:
I added a link from the main page of the docs back to the repo. This can be very helpful but isn't included in Documenter.jl right now, see also https://discourse.julialang.org/t/link-back-to-main-repo-on-each-page-in-documenter-jl/57444.

By the way, thanks for this nice package!


ranocha:
Bump. Any interest?

PR #187, opened 2021-10-10
fix cdot bug for complicated expressions

ranocha:
`latexoperation` removed the keyword argument `cdot` before.


gustaphe:
The alternative is to keep it, but add those in the inner function call:

```julia
function outer(args; special=true, kwargs...)
    return inner(args; kwargs..., special)
end
```

ranocha:
Bump. Any interest? It would be really great to get this merged

korsbo:
This makes sense, thanks!

PR #186, opened 2021-10-07
add SymPy plugin

ranocha:
I hope this follows the recommended style. When it works, it would be really great to get a new release of Latexify.jl.

PR #188, opened 2021-10-11
improve documentation of at-latexrecipe

ranocha:
This basically came up after several failures from my side, see https://github.com/korsbo/Latexify.jl/pull/186#issuecomment-939472965


korsbo:
Great, thanks! 

I'll see if I can filter my GH notifications better so that PRs like these don't slip by unnoticed - I'm sorry about the wait.

Issue #197, opened 2021-12-20
Failing sqrt symbol


t-bltg:
`latexify(:(√π))`'s generated `tex` fails to compile:

```julia
julia> using Latexify
julia> latexify(:(√π)) |>
x -> raw"\documentclass{standalone}\begin{document}" * x * raw"\end{document}" |>
x -> run(pipeline(`pdflatex`, stdin=IOBuffer(x)))

This is pdfTeX, Version 3.141592653-2.6-1.40.22 (TeX Live 2021) (preloaded format=pdflatex)
 restricted \write18 enabled.
**entering extended mode
LaTeX2e <2021-11-15>
L3 programming layer <2021-11-22>
[...]
! Missing { inserted.
<to be read again> 
                   \left 
<*> ...lass{standalone}\begin{document}$\sqrt\left
                                                  ( \pi \right)$\end{document}
? 
! Emergency stop.
<to be read again> 
                   \left 
<*> ...lass{standalone}\begin{document}$\sqrt\left
                                                  ( \pi \right)$\end{document}
!  ==> Fatal error occurred, no output PDF file produced!
Transcript written on texput.log.
ERROR: LoadError: failed process: Process(`pdflatex`, ProcessExited(1)) [1]

Stacktrace:
 [1] pipeline_error
   @ ./process.jl:531 [inlined]
 [2] run(::Base.CmdRedirect; wait::Bool)
   @ Base ./process.jl:446
 [3] run
   @ ./process.jl:444 [inlined]
 [4] #6
   @ ./REPL[2]:1 [inlined]
 [5] |>
   @ ./operators.jl:966 [inlined]
 [6] #5
   @ ./REPL[2]:1 [inlined]
 [7] |>(x::LaTeXStrings.LaTeXString, f::var"#5#7")
   @ Base ./operators.jl:966
 [8] top-level scope
   @ REPL[2]:1
```

PR #198, opened 2021-12-20
Fix sqrt construct

t-bltg:
Fix https://github.com/korsbo/Latexify.jl/issues/197, added tests.

Are there any other similar constructs similar to `(sqrt, √) ` requiring this kind of fixes ?

EDIT: also added `∛` and `∜` from https://docs.julialang.org/en/v1/manual/unicode-input/.


korsbo:
Looks good, thanks 😄

PR #202, opened 2022-01-06
Change MIME type `image/pdf` to `application/pdf`

MarkusLohmayer:
I suppose this is a tiny bug fix:
The fall-back MIME type (if none of the priority MIME types are `displayable`) was set to `image/pdf` which does not seem to be a correct MIME type. Therefore just calling `render` did not work for me, whereas specifying `application/pdf` explicitly did the job. Thanks.


gustaphe:
Thank you! This looks clear-cut to me.

PR #205, opened 2022-01-31
Add sum and prod

gustaphe:
This operation had to happen ahead of the recursion, but I think the result is pretty neat:

test

```julia
latexalign(fill(:(A), 4), [
                              :(prod(x_n)),
                              :(sum(x_n for n in :)),
                              :(prod(x_n for n in N)),
                              :(sum(x_n for n in n_0:N)),
                             ])
```


korsbo:
Very cool, nice one!

PR #207, opened 2022-02-15
Rework macros

gustaphe:
The macros (`@latexrun`, `@latexify`) were going via strings, which meant they ignored some of the latexify logic. It also meant you couldn't run them using an interpolated value (variable, function call, ...).

With this PR, the macros are operating directly on the input expressions (with some escaping logic). The star of the show is you can now do 

```julia
julia> using Latexify, Unitful, UnitfulLatexify
julia> @latexrun v = 3.5u"m/s" # This has not changed
L"$v = 3.5 \cdot \mathrm{@u\_str}\left( m/s \right)$"
julia> @latexrun v = $(3.5u"m/s") # would error before the PR
L"$v = 3.5\;\mathrm{m}\,\mathrm{s}^{-1}$"
```

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -179,6 +179,16 @@ They can for example be useful for latexifying simple mathsy functions like
 lstr = @latexrun f(x; y=2) = x/y
 ```
 
+The arguments to the macro can be interpolated with `$` to use the actual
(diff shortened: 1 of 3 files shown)


korsbo:
Cool, nice job!

Issue #206, opened 2022-02-14
UTF modifiers causing latexification issues


finmod:
@korsbo The issue here is that  variables written in Latex as `$ \boldsymbol{\omega}(t), \boldsymbol{\lambda}(t), \mathbf{d}(t), \mathbf{y}_{\boldsymbol{e}}(t), \mathbf{u}(t), \mathbf{y}_{\boldsymbol{d}}(t) $` are not rendered accurately in MTK coding and `equations()`. In particular, letter subscripts are not supported whereas numeral subscripts are. 

Also, there is the question of  a variable `$\boldsymbol{\pi}_{\boldsymbol{e} }$` not being confused with the numerical `pie`

korsbo:
closed by #208

finmod:
There remains a small issue with subscripted variables. In printing out this example, you can see that subscripted quantities are represented properly some of the times but not all the time (e.g. y_d and y_e)
```julia
 ModelingToolkit.@parameters begin
      
      η=500
      C=0.33
      b=0.135
      α=0.025
      β=0.02
      δ=0.05
      r=0.03
  
      ϕ₀=0.0401
      ϕ₁=6.41e-05
  
      k₀=-0.0065
      k₁=-5.0
      k₂=20
      ηu=0.2
      u_b=0.9 
  
      m=1.2
      γ=0.4
      f_d=0.1
      
      η₁=0.8
      η₂=0.2
      η₃=0.7
      ηₑ=0.8 
      
      θ₁=0.9 
      θ₂=0.8 
      θ₃=-5.0
      e₁=0.0
      g₂= α + β
end    

@variables begin
      t
      ω(t) = 0.9
      λ(t) = 0.9
      d(t) = 0.5
      y_e(t) = 0.9
      u(t) = 0.8
      y_d(t) = 0.9
      
      ν(t) = 3.0
      φ(t)
      ι(t)
      κ(t)
      p_e(t)
      θ(t) 
      g(t) 
end

Dₜ = Differential(t)
tmax = 25.0                         # 25 years


# Behavioural equations

p_e = (t, ω, y_e, d) ->           y_e*(1 - ω) - r*d
ν  = (t, ω) ->                    (1/C)*((1 -  ω)/b)^(-1/η)
φ  = (t, λ) ->                    -ϕ₀ + ϕ₁/((1 - λ)^2)
[...]

PR #208, opened 2022-02-15
Fix some regular unicode conversions

gustaphe:
#206 pointed to some unicode2latex substitutions which could be fixed with some simple (thinking-wise, at least) find-and-replace operations.

Some of these only work with specific latex packages, like `unicode-math`, but that's the only way things like `𝝙` were ever going to be rendered anyway.


korsbo:
Great, thanks!

PR #212, opened 2022-02-24
Add @latexdefine macro

gustaphe:
Adds a third printing macro `@latexdefine`, which does both of the things that `@latexrun` does (prints and evaluates expression), but also adds an equals sign and the evaluated value of the expression. This is useful for things like a Pluto workflow, where you can then in a single line get a variable's (new) definition.

```julia
julia> x = 35;

julia> @latexdefine y = 2x
L"$y = 2 \cdot x = 70$"

julia> @latexdefine y
L"$y = 70$"
```

I'm not married to the name.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1,3 +1,39 @@
+"""
+    @latexify expression
+
+Create `LaTeXString` representing `expression`.
+Variables and expressions can be interpolated with `\$`.
+
+# Examples
+```julia-repl
+julia> @latexify x^2 + 3/2
+L"\$x^{2} + \\frac{3}{2}\$"
+
+julia> @latexify x^2 + \$(3/2)
+L"\$x^{2} + 1.5\$"
+```
+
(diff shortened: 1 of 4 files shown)


src/macros.jl
```diff
@@ -13,6 +49,35 @@ macro latexrun(expr)
...
+"""
+macro latexdefine(expr)
+    return esc(
+        :(latexify(
+            Expr(
+                 :(=),
+                $(Meta.quot(expr)),
+                $(postwalk(expr) do ex
+                    if ex isa Expr && ex.head == :$
+                        return ex.args[1]
```
korsbo, line 76: Could you please add a test that hits these lines?

korsbo (changes requested):
Looks good! You're not testing the interpolation machinery for `@latexdefine` though. If you add that then you're good to go :)

korsbo (approved):

korsbo:
🎉

PR #214, opened 2022-03-15
Macrokwargs

gustaphe:
Allow keyword arguments to be passed to the latexify macros:

```julia
julia> @latexify x = 3 env=:raw
L"x = 3"
```


gustaphe:
Hm, looks like running the `runtests` file at all fails on versions before the bare keyword syntax. Might just remove that test, can't see a way around it (and have a hard time testing it without installing an old Julia version).

korsbo (approved):

PR #218, opened 2022-03-29
Adjustment

gustaphe:
Allow vectors of adjustment symbols, and make adjustment affect mdtable.


gustaphe:
This way was cleaner if it passes CI. Otherwise I might just do that.

PR #219, opened 2022-04-21
Function names with subscripts

### LINKED ISSUE #194: Function names with subscripts
The way subscripts are handled seems inconsistent between "functions" and "variables".

For example, 
```
julia> @latexrun f_a = π
L"$f_{a} = \pi$"
```
represents it in LaTeX as a subscripted variable. On the other hand, defining a function with the same syntax

```julia
julia> @latexrun f_b(x) = 2 * x
L"$\mathrm{f\_b}\left( x \right) = 2 \cdot x$"
```
escapes the underscore, such that it remains an underscore in the LaTeX representation. Is this difference between variables and functions on purpose?

gustaphe:
Closes #194 

Bikeshed moment: Currently it does `f_abc() => f_{\mathrm{abc}}()`, but `f_\mathrm{abc}()` and `f_{abc}()` could both be acceptable to me. Any opinions, @korsbo ?

I don't really like that I reuse `convert_subscripts` for this, it's a bit opaque, but it would have rewritten a lot of the code otherwise. Perhaps the function should have another name.


gustaphe:
Also, maybe long variable names should also be mathrm'd?

korsbo:
This is one of those places where programming and maths notation clash and there's no good default. I think I'd prefer `f_{abc}()` in the example you posted above.  `\mathrm` makes sense to me when used on an entire name of an object but I can't quite remember why we'd be mixing maths and text mode like that. 

As far as I can tell, Latexify is now mostly used to produce proper maths notation and not the mixed programming language + maths stuff that I initially used it for. Maybe it's about time to just make maths mode the default in all that we do and just have some kwarg that toggles snake case.

gustaphe:
> some kwarg

Way ahead of you (as of #215):

```julia
julia> @latexify something_else(x) snakecase=true
L"$something\_else\left( x \right)$"
```

gustaphe:
I think there's room for a future PR to  make multi-character variables, subscripts etc romanized (by kwarg or default or whatever), but this (`f_{abc}`) seems like an acceptable low-impact choice for now.

The argument *for* romanizing long subscripts is that -- at least in my experience -- things like `f_\mathrm{max}` is fairly common, and you can always go `f_{m,a,x}` if you mean those as separate variables (which that future PR would need to take care of ...)

Anyway, how's this?

PR #220, opened 2022-04-28
Add  kwarg post

gustaphe:
Adds a `post` kwarg to `@latexdefine`, which gives you a bit more control of how the final rhs is formatted.

```julia
julia> @latexdefine x=3.75 post=x->round(x; sigdigits=2)
L"$x = 3.75 = 3.8$"

julia> @latexdefine x=1//2 post=float
L"$x = \frac{1}{2} = 0.5$"

julia> using Unitful, UnitfulLatexify; m = 5u"kg"; v = 10u"m/s";

julia> @latexdefine E = m*v^2/2 post=u"J"
L"$E = \frac{m \cdot v^{2}}{2} = 250.0\;\mathrm{J}$"
```

I'm mostly interested in this last one for physics material, but I think this is general enough it could be useful in different contexts as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '0.7' 
           - '1' # Leave this line unchanged. '1' will automatically expand to the latest stable 1.x release of Julia.
           - 'nightly'
         os:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -15,7 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '0.7'
```
gustaphe, line 18: Okay I'm doing it.

korsbo: Could you then please bump the compat bound in Project.toml and also the latexify minor version?

korsbo: nah, not in that case

Issue #182, opened 2021-08-19
Doesn't translate accents


BeastyBlacksmith:
```julia
julia> latexify("ç")
L"$ç$"

julia> latexify("ã")
L"$ã$"
```
should become

```julia
julia> latexify("ç")
L"\c{c}"

julia> latexify("ã")
L"\~{a}"
```

I'd say.

korsbo:
Hmmm, it seems like the massive unicode -> latex dict in `src/unicode2latex.jl` is incomplete. That's a bit troubling since I have no desire to manually curate such a mapping. Any individual mappings could easily be PRd by just adding entries, but the deeper problem is why there are any such mappings missed at all.

PR #223, opened 2022-07-07
Support diacritics (accents, ...)

t-bltg:
- handle diacritics in a consistent manner, using escape sequences from https://en.wikibooks.org/wiki/LaTeX/Special_Characters#Escaped_codes.

```julia
julia> using Latexify
julia> latexify("éçà")
L"$\textrm{\'{e}}\textrm{\c{c}}\textrm{\`{a}}$"
```
- support `QuoteNode` in `latexraw` (e.g. when parsing `"Plots.jl"` as expr, corner case);
- add `latexify` docstring for `help?> latexify`;
- add failing expr error message when parsing fails;
- added a `parse` keyword to `_latexraw`, since `Meta.parse` can fail (especially parsing string containing diacritric markers - strings are normalized), and dispatch on `Val{true/false}` to `Meta.parse` or not.

Fix https://github.com/korsbo/Latexify.jl/issues/182.
Goes with https://github.com/JuliaPlots/Plots.jl/pull/4262.

### DIFF
--- a/src/latexraw.jl
+++ b/src/latexraw.jl
@@ -157,7 +157,7 @@ function _latexraw(::Val{true}, i::String; kwargs...)
         error("""
 in Latexify.jl:
 You are trying to create latex-maths from a `String` that cannot be parsed as
-an expression.
+an expression: `$i`.
 
 `latexify` will, by default, try to parse any string inputs into expressions
 and this parsing has just failed.
--- a/src/latexraw.jl
+++ b/src/latexraw.jl
@@ -59,7 +59,7 @@ latexraw(args...; kwargs...) = process_latexify(args...; kwargs..., env=:raw)
 
 function _latexraw(inputex::Expr; convert_unicode=true, kwargs...)
     ## Pass all arrays or matrices in the expr to latexarray
-    inputex = postwalk(x -> x isa Expr && x.head in [:hcat, :vcat, :vect, :typed_vcat, :typed_hcat] ?
+    inputex = postwalk(x -> Meta.isexpr(x, [:hcat, :vcat, :vect, :typed_vcat, :typed_hcat]) ?
                        latexarray(expr_to_array(x); kwargs...)
                        : x,
                        inputex)
@@ -68,12 +68,12 @@ function _latexraw(inputex::Expr; convert_unicode=true, kwargs...)
     function recurseexp!(ex)
         prevOp = Vector{Symbol}(undef, length(ex.args))
         fill!(prevOp, :none)
-        if ex.head==:call && ex.args[1] in (:sum, :prod) && ex.args[2] isa Expr && ex.args[2].head == :generator
+        if Meta.isexpr(ex, :call) && ex.args[1] in (:sum, :prod) && Meta.isexpr(ex.args[2], :generator)
             op = ex.args[1]
             term = latexraw(ex.args[2].args[1])
             gen = ex.args[2].args[2]
             itervar = latexraw(gen.args[1])
-            if gen.args[2] isa Expr && gen.args[2].head == :call && gen.args[2].args[1] == :(:)
+            if Meta.isexpr(gen.args[2], :call) && gen.args[2].args[1] == :(:)
                 # sum(x_n for n in n_0:N) => \sum_{n=n_0}^{N} x_n
                 lower = latexraw(gen.args[2].args[2])
                 upper = latexraw(gen.args[2].args[end])
--- a/src/latexraw.jl
+++ b/src/latexraw.jl
@@ -111,9 +111,11 @@ function _latexraw(args...; kwargs...)
 end
 _latexraw(arr::Union{AbstractArray, Tuple}; kwargs...) = _latexarray(arr; kwargs...)
 _latexraw(i::Nothing; kwargs...) = ""
-_latexraw(i::SubString; kwargs...) = latexraw(Meta.parse(i); kwargs...)
+_latexraw(i::SubString; parse=true, kwargs...) = latexraw(parse ? Meta.parse(i) : i; kwargs...)
 _latexraw(i::SubString{LaTeXStrings.LaTeXString}; kwargs...) = i
 _latexraw(i::Rational; kwargs...) = i.den == 1 ? latexraw(i.num; kwargs...) : latexraw(:($(i.num)/$(i.den)); kwargs...)
+_latexraw(i::QuoteNode; kwargs...) = _latexraw(i.value)
+
 function _latexraw(z::Complex; kwargs...)
     if iszero(z.re)
         isone(z.im) && return LaTeXString(get(kwargs, :imaginary_unit, "\\mathit{i}"))
@@ -142,7 +144,12 @@ function _latexraw(i::Symbol; convert_unicode=true, snakecase=false, safescripts
     return LaTeXString(str)
 end
 
-function _latexraw(i::String; kwargs...)
+_latexraw(i::String; parse=true, kwargs...) = _latexraw(Val(parse), i; kwargs...)
+
+_latexraw(::Val{false}, i::String; convert_unicode=true, kwargs...) =
+    LaTeXString(convert_unicode ? unicode2latex(i) : i)
+
+function _latexraw(::Val{true}, i::String; kwargs...)
     try
         ex = Meta.parse(i)
         return latexraw(ex; kwargs...)
--- a/src/unicode2latex.jl
+++ b/src/unicode2latex.jl
@@ -21,7 +21,6 @@ import Base.Unicode
 """
 function latex_diacritics(c::Char)
     c = lowercase(c)
-    @assert length(c) == 1
     out = []
     for p in (
         # en.wikibooks.org/wiki/LaTeX/Special_Characters#Escaped_codes
--- a/src/unicode2latex.jl
+++ b/src/unicode2latex.jl
@@ -1,4 +1,60 @@
-const unicodedict = Dict{Char, String}(
+import Base.Unicode
+
+"""
+- generate latex escape codes for diacritics of the latin alphabet (see https://en.wikibooks.org/wiki/LaTeX/Special_Characters#Escaped_codes)
+- `@upper_lower` is used to generate a subset of the following sequence (uppercase and lowercase characters):
+    - 'à' => "\\`{a}"  # grave
+    - 'á' => "\\'{a}"  # acute
+    - 'ä' => "\\"{a}"  # umlaut (trema, dieresis)
+    - 'a̋' => "\\H{a}"  # hungarian umlaut (double acute)
+    - 'ã' => "\\~{a}"  # tilde
+    - 'â' => "\\^{a}"  # circumflex
+    - 'a̧' => "\\c{a}"  # cedilla
+    - 'ą' => "\\k{a}"  # ogonek
+    - 'ā' => "\\={a}"  # macron (bar above)
+    - 'a̱' => "\\b{a}"  # bar under
+    - 'ȧ' => "\\.{a}"  # dot above
+    - 'ạ' => "\\d{a}"  # dot under
+    - 'å' => "\\r{a}"  # ring
+    - 'ă' => "\\u{a}"  # breve
+    - 'ǎ' => "\\v{a}"  # caron (háček)
+"""
+function latex_diacritics(c::Char)
+    c = lowercase(c)
+    @assert length(c) == 1
+    out = []
+    for p in (
+        # en.wikibooks.org/wiki/LaTeX/Special_Characters#Escaped_codes
+        '`' => 0x300,  # latex \`{c} maps to 'c' * Char(0x300)
+        "'" => 0x301,
+        '^' => 0x302,
+        '~' => 0x303,
+        '=' => 0x304,
+        'u' => 0x306,
+        '.' => 0x307,
+        '"' => 0x308,
+        'r' => 0x30a,
+        'H' => 0x30b,
+        'v' => 0x30c,
+        'd' => 0x323,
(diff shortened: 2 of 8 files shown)


gustaphe:
Hi! Thank you for the contribution, this looks like a good idea, but I haven't done a proper review yet.

However, I feel the need to comment on your workflow. You seem to be force-pushing a *lot*. That's a pretty extreme measure, and makes it difficult to follow your history. Please make successive commits rather than amending a single commit and force-pushing it. I'm certain most repos will agree with this.

t-bltg:
> Hi! Thank you for the contribution, this looks like a good idea, but I haven't done a proper review yet.

Thanks, I think this can be reviewed now.

>  You seem to be force-pushing a lot. That's a pretty extreme measure, and makes it difficult to follow your history.

Pushing a lot of commits is certainly a bad idea too. I can rebase this PR though, to make things clearer.

gustaphe:
I would certainly prefer several commits over a constantly changing history. If your want to cause fewer commits, consider simply pushing less often.

t-bltg:
Rebased for clarity. The last commit is optional and can be removed if required.

t-bltg:
@gustaphe, would it be possible to review (and merge this PR / bump ver iff approved) so that it does not block https://github.com/JuliaPlots/Plots.jl/pull/4262 ?

src/unicode2latex.jl
```diff
@@ -1,4 +1,60 @@
...
+        'k' => 0x328,
+        'b' => 0x331,
+    )
+        latex_escape, mark = p.first, Char(p.second)
+        lower, upper = c * mark, uppercase(c) * mark
+        # e.g. ('y' * Char(0x30A) == "ẙ") != (Char(0x1e99) == 'ẙ')
+        for p in (lower => "\\textrm{\\$latex_escape{$c}}", upper => "\\textrm{\\$latex_escape{$(uppercase(c))}}")
+            push!(out, p)
+            if (alias = length(p.first) == 1 ? p.first : Unicode.normalize(p.first)) != p.first
+                push!(out, (length(alias) == 1 ? first(alias) : alias) => p.second)
```
gustaphe, line 50: Can `length(alias) != 1` here? In what cases, and is it then not a problem to only keep `first(alias)`?

src/unicode2latex.jl
```diff
@@ -1,4 +1,60 @@
...
+            end
+        end
+    end
+    out
+end
+
+const unicodedict = Dict{Union{Char,String}, String}(
```
gustaphe, line 57: How much simpler would the code get if you just made this a `Dict{String, String}`? Latexify should rarely be a performance bottleneck, and I'm not even sure `Dict{Union{Char, String}, String}` is better performance wise.

src/unicode2latex.jl
```diff
@@ -2450,26 +2506,39 @@ const unicodedict = Dict{Char, String}(
     'Χ' => raw"\Chi",
     '⤮' => raw"\neovsearrow",
     '•' => raw"\bullet",
-    )
+    (latex_diacritics.('a':'z')...)...,
```
gustaphe, line 2509: Is this not the same thing as just 
```suggestion
    latex_diacritics.('a':'z')...,
```
?

test/unicode2latex.jl
```diff
@@ -12,3 +12,27 @@ raw"\begin{equation}
 ", "\r\n"=>"\n")
 
 @test latexify("αaβ") == raw"${\alpha}a\beta$"
+
+@test latexify("αaβ").s == raw"${\alpha}a\beta$"
+
+@test latexify("ÀéÜ"; parse=false).s == raw"$\textrm{\`{A}}\textrm{\'{e}}\textrm{\\\"{U}}$"
```
gustaphe, line 18: Is there a reason for the extra `\\` with diaereses?

src/latexraw.jl
```diff
@@ -131,26 +139,30 @@ function _latexraw(i::Number; fmt=PlainNumberFormatter(), kwargs...)
...
 function _latexraw(i::Symbol; convert_unicode=true, snakecase=false, safescripts=false, kwargs...)
-    str = string(i == :Inf ? :∞ : i)
+    str = string(i === :Inf ? :∞ : i)
     str = convert_subscript(str; snakecase=snakecase)
     convert_unicode && (str = unicode2latex(str; safescripts=safescripts))
     return LaTeXString(str)
 end
 
-function _latexraw(i::String; kwargs...)
+_latexraw(i::String; parse=true, kwargs...) = _latexraw(Val(parse), i; kwargs...)
```
gustaphe, line 152: Any strong motivation to go to a `Val{Bool}` here? Seems like this could just be a single `if` with an early return, which would leave the code more self-explanatory and so easier to maintain.

gustaphe (commented):
I disagree with the choice of making all of these `\textrm`. If a character in math mode is supposed to be romanized, then that should be a property of the variable it signifies, not its spelling. To me it's better to get an `Command \" invalid in math mode` error than have it silently put some characters in roman.

src/unicode2latex.jl
```diff
@@ -1,4 +1,60 @@
...
+        'c' => 0x327,
+        'k' => 0x328,
+        'b' => 0x331,
+    )
+        latex_escape, mark = p.first, Char(p.second)
+        lower, upper = c * mark, uppercase(c) * mark
+        # e.g. ('y' * Char(0x30A) == "ẙ") != (Char(0x1e99) == 'ẙ')
+        for p in (lower => "\\textrm{\\$latex_escape{$c}}", upper => "\\textrm{\\$latex_escape{$(uppercase(c))}}")
+            push!(out, p)
+            if (alias = length(p.first) == 1 ? p.first : Unicode.normalize(p.first)) != p.first
```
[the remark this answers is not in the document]
t-bltg: I'll split that.

src/unicode2latex.jl
t-bltg: I didn't want to throw too many changes with the conversion of char key to string, but it's probably better.

src/latexraw.jl
```diff
@@ -59,7 +59,7 @@ latexraw(args...; kwargs...) = process_latexify(args...; kwargs..., env=:raw)
 
 function _latexraw(inputex::Expr; convert_unicode=true, kwargs...)
     ## Pass all arrays or matrices in the expr to latexarray
-    inputex = postwalk(x -> x isa Expr && x.head in [:hcat, :vcat, :vect, :typed_vcat, :typed_hcat] ?
+    inputex = postwalk(x -> x isa Expr && Meta.isexpr(x, [:hcat, :vcat, :vect, :typed_vcat, :typed_hcat]) ?
```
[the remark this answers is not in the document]
t-bltg: Yes, `Meta` is not exported. Do you want these changes back ?

src/latexraw.jl
```diff
@@ -122,35 +125,39 @@ function _latexraw(z::Complex; kwargs...)
...
 function _latexraw(i::Symbol; convert_unicode=true, snakecase=false, safescripts=false, kwargs...)
-    str = string(i == :Inf ? :∞ : i)
+    str = string(i === :Inf ? :∞ : i)
     str = convert_subscript(str; snakecase=snakecase)
     convert_unicode && (str = unicode2latex(str; safescripts=safescripts))
     return LaTeXString(str)
 end
 
-function _latexraw(i::String; kwargs...)
+_latexraw(i::String; parse=true, kwargs...) = _latexraw(Val(parse), i; kwargs...)
```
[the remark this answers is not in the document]
t-bltg: All right, I'm going to drop the last format commit.

src/unicode2latex.jl
t-bltg: No, because you have to splat (unpack) all diacritics for a single char, and then do that for all chars in the range `'a':'z'`.

src/latexraw.jl
t-bltg: Well I always prefer to dispatch on types, especially since the `_latexraw(Val{true}, ...)` uses `try catch` which makes the function unpure (side effects).

src/unicode2latex.jl
t-bltg: This makes tests fail in subtle ways, so I'm going to revert that.

src/unicode2latex.jl
t-bltg: Yes, e.g. `length("ẙ") == 2` it's the character `y` followed by the diacritic marker `Char(0x30a)`. And this is different from `'ẙ'` which is encoded as a single character. And not all combination of diacritics have a single character equivalent. Unicode is subtle.

test/unicode2latex.jl
t-bltg: Yes, you have to correctly escape `"` which is embedded in a string.

src/latexraw.jl
```diff
@@ -104,16 +104,19 @@ function _latexraw(inputex::Expr; convert_unicode=true, kwargs...)
...
-
 function _latexraw(args...; kwargs...)
     @assert length(args) > 1 "latexify does not support objects of type $(typeof(args[1]))."
     _latexraw(args; kwargs...)
 end
 _latexraw(arr::Union{AbstractArray, Tuple}; kwargs...) = _latexarray(arr; kwargs...)
 _latexraw(i::Nothing; kwargs...) = ""
-_latexraw(i::SubString; kwargs...) = latexraw(Meta.parse(i); kwargs...)
-_latexraw(i::SubString{LaTeXStrings.LaTeXString}; kwargs...) = i
+_latexraw(i::SubString; parse=true, kwargs...) = latexraw(Val(parse), i; kwargs...)
```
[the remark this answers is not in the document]
t-bltg: Yes, I've removed it.

t-bltg:
> I disagree with the choice of making all of these \textrm. If a character in math mode is supposed to be romanized, then that should be a property of the variable it signifies, not its spelling. To me it's better to get an Command \" invalid in math mode error than have it silently put some characters in roman.

I'm not sure I'm following you here: the point of `latexify` is to render an equation isn't it ? The only way (if I'm not misunderstanding, and without loading extra packages) to have latex compile accentuated characters in equations is to use `\textrm` ?

src/latexraw.jl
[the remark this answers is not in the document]
t-bltg: Restored.

src/unicode2latex.jl
```diff
@@ -1,4 +1,60 @@
...
+    - 'a̱' => "\\b{a}"  # bar under
+    - 'ȧ' => "\\.{a}"  # dot above
+    - 'ạ' => "\\d{a}"  # dot under
+    - 'å' => "\\r{a}"  # ring
+    - 'ă' => "\\u{a}"  # breve
+    - 'ǎ' => "\\v{a}"  # caron (háček)
+"""
+function latex_diacritics(c::Char)
+    c = lowercase(c)
+    @assert length(c) == 1
```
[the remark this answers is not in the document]
t-bltg: Removed.

gustaphe:
That may be right. My objection is exemplified in this hypothetical: say I define a type

```julia
struct TextSub
var::String
sub::String
end

@latexrecipe function f(x::TextSub)
    return Expr(:latexifymerge, 
        x.var, 
        "_\textrm{", 
        x.sub, 
        "}"
    )
end
```
(Possible misspellings, I'm on my phone traveling)

This is how I would use anything where I expected romanized text, and so anything where I would expect `ä` to print correctly. But if some characters are *always* surrounded by `textrm`, this becomes difficult to do.

There is precedent for using kwargs to introduce formatting commands on demand, could that be the solution here?

That said, maybe this is an inconsequential objection. I'll survive.

t-bltg:
I don't understand how your example shows a valid point since it doesn't compile: `latexifymerge` returns an equation, so escape sequence are invalid in this tex mode.

```julia
using Latexify

struct TextSub
  var::String
  sub::String
end

@latexrecipe f(x::TextSub) = return Expr(:latexifymerge, x.var, raw"_\textrm{", x.sub, "}")

main() = begin
  fn = tempname()
  write(
    fn, """
    \\documentclass{standalone}
    \\begin{document}
      $(latexify(:($(TextSub("é", "ç")) + $(TextSub("c", "123")))))
    \\end{document}
    """
  )
  run(`latexmk -quiet -pdf $fn`)
  return
end

main()
```

**with PR**
```
Rc files read:
  NONE
Latexmk: Run number 1 of rule 'pdflatex'
This is pdfTeX, Version 3.141592653-2.6-1.40.24 (TeX Live 2022) (preloaded format=pdflatex)
 restricted \write18 enabled.
entering extended mode
Latexmk: Getting log file 'jl_qMW3dH.log'
  0.701773 seconds (3.62 k allocations: 267.109 KiB)
```

**without PR**
```
Rc files read:
  NONE
Latexmk: Run number 1 of rule 'pdflatex'
This is pdfTeX, Version 3.141592653-2.6-1.40.24 (TeX Live 2022) (preloaded format=pdflatex)
 restricted \write18 enabled.
entering extended mode
Latexmk: Getting log file 'jl_WXLw8J.log'
Collected error summary (may duplicate other messages):
  pdflatex: Command for 'pdflatex' gave return code 1
      Refer to 'jl_WXLw8J.log' for details
Latexmk: If appropriate, the -f option can be used to get latexmk
  to try to force complete processing.
ERROR: failed process: Process(`latexmk -quiet -pdf /tmp/jl_WXLw8J`, ProcessExited(12)) [12]
```

t-bltg:
@korsbo or @gustaphe, if no more objections, would it be possible to go forward on this PR ?

korsbo (approved):

t-bltg:
Thanks ! Don't hesitate to ping me if any issue pops up related to this PR.

PR #229, opened 2022-09-10
update docs

t-bltg:
Following https://github.com/korsbo/Latexify.jl/pull/225.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -237,7 +237,7 @@ Alternatively, `render(str, mime)` can also be used to generate and display DVI,
 latexify(:(x/y)) |> s -> render(s, MIME("image/png"))
 ```
 
-PNG output relies on [ghostscript](https://www.ghostscript.com) or alternatively on [dvipng](http://www.nongnu.org/dvipng) (use the `convert` argument).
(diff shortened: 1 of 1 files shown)


docs/src/index.md
```diff
@@ -237,18 +237,26 @@ Alternatively, `render(str, mime)` can also be used to generate and display DVI,
 latexify(:(x/y)) |> s -> render(s, MIME("image/png"))
 ```
 
-PNG and SVG outputs rely on [dvipng](http://www.nongnu.org/dvipng/) and [dvisvgm](https://dvisvgm.de/), respectively.
+PNG output relies on [ghostscript](https://www.ghostscript.com) or alternatively on [dvipng](http://www.nongnu.org/dvipng) (use the `convert` argument).
```
gustaphe, line 240: What determines the "alternatively" part? What does "use the `convert` argument" refer to?

t-bltg: https://github.com/korsbo/Latexify.jl/blob/956eccb5982244649087c909f2895400b03d6cb5/src/utils.jl#L154

and 

https://github.com/korsbo/Latexify.jl/blob/956eccb5982244649087c909f2895400b03d6cb5/src/utils.jl#L190

t-bltg: Maybe I should add: use the _`convert` keyword argument of the function `render` instead_ ?

gustaphe: Yes, I'm thinking from the perspective of someone reading the documentation. Perhaps something like "if `render` is called with `convert=:gs` then ..."

korsbo (approved):

PR #247, opened 2023-02-03
BenchmarkCI

gustaphe:
An idea after #244 . I don't see a way to have it report loading time, sadly no loading time report. And we probably want to be running a bigger array of benchmarks, but I wanted to try it out.


gustaphe:
I would have hoped it could include a report even if it could only benchmark the current version, but sadly. It's angry that there's no `benchmarks` folder on master, so it cannot compare.

Issue #256, opened 2023-04-21
Calling render with name argument that does not include a directory throws ERROR: IOError: cd(""): no such file or directory (ENOENT)


reallyasi9:
In answering [this StackOverflow question](https://stackoverflow.com/questions/76006842/how-to-save-a-latex-rendered-image-via-latexify-to-disk-in-png-in-julia/76009751), we discovered that `render`, when called with a `name` argument that does not include a directory, will throw. The culprit is line 105 of `utils.jl`, which does not check if `dirname(name)` is empty before attempting to change directory.

https://github.com/korsbo/Latexify.jl/blob/956eccb5982244649087c909f2895400b03d6cb5/src/utils.jl#L105

A reasonable fix might be:

```julia
dir = dirname(name)
if isempty(dir)
    dir = "."
end
cd(dir) do
    ...
end
```

gustaphe:
Good catch. In fact, I think it should be separated, 
```julia
if isnothing(name)
    name = replace(source_name, ".tex"=>".png") # or whatever the robust way to replace an extension is
else
    mv(replace(source_name, ".tex"=>".png"), name)
end
```

That way you don't clutter the directory with aux files. Everything is generated in a temp directory, and the final result is moved to where you want it. I could write something up this weekend, or do you want to?

gustaphe:
Closed in #257

PR #257, opened 2023-04-23
Less file clutter when rendering

gustaphe:


gustaphe:
Closes #256 , also creates a script for the README images, to create them more consistently.

Issue #258, opened 2023-04-23
Error requiring `DiffEqBase` from `Latexify`


cadojo:
Using `ModelingToolkit` and/or `Symbolics` results in an error, seemingly while importing `Latexify`. This happens when I `]add AstrodynamicalModels#main`, then enter `using AstrodynamicalModels`.

<details>

```
┌ Warning: Error requiring `DiffEqBase` from `Latexify`
│   exception =
│    LoadError: UndefVarError: ParameterizedFunctions not defined
│    Stacktrace:
│      [1] top-level scope
│        @ ~/.julia/packages/Latexify/L5vkp/ext/ParameterizedFunctionsExt.jl:4
│      [2] include(mod::Module, _path::String)
│        @ Base ./Base.jl:419
│      [3] include(x::String)
│        @ Latexify ~/.julia/packages/Latexify/L5vkp/src/Latexify.jl:1
│      [4] macro expansion
│        @ ~/.julia/packages/Requires/Z8rfN/src/Requires.jl:40 [inlined]
│      [5] top-level scope
│        @ ~/.julia/packages/Latexify/L5vkp/src/Latexify.jl:69
│      [6] eval
│        @ ./boot.jl:368 [inlined]
│      [7] eval
│        @ ~/.julia/packages/Latexify/L5vkp/src/Latexify.jl:1 [inlined]
│      [8] (::Latexify.var"#179#191")()
│        @ Latexify ~/.julia/packages/Requires/Z8rfN/src/require.jl:101
│      [9] macro expansion
│        @ timing.jl:382 [inlined]
│     [10] err(f::Any, listener::Module, modname::String, file::String, line::Any)
│        @ Requires ~/.julia/packages/Requires/Z8rfN/src/require.jl:47
│     [11] (::Latexify.var"#178#190")()
│        @ Latexify ~/.julia/packages/Requires/Z8rfN/src/require.jl:100
│     [12] withpath(f::Any, path::String)
│        @ Requires ~/.julia/packages/Requires/Z8rfN/src/require.jl:37
│     [13] (::Latexify.var"#177#189")()
│        @ Latexify ~/.julia/packages/Requires/Z8rfN/src/require.jl:99
│     [14] #invokelatest#2
│        @ ./essentials.jl:729 [inlined]
│     [15] invokelatest
│        @ ./essentials.jl:726 [inlined]
│     [16] foreach(f::typeof(Base.invokelatest), itr::Vector{Function})
[...]

PR #259, opened 2023-04-24
Remove deprecated extensions code

ChrisRackauckas:
These have been deprecated for a long time. The ParameterizedFunction one doesn't need to exist anyways because it's just MTK since 2020. So these can just be deleted and that fixes issues like https://github.com/korsbo/Latexify.jl/issues/258


gustaphe:
I haven't wanted to touch this because I don't know the involved packages. How deprecated are they? Would this be breaking to anyone?

I know these packages were a large part of why @korsbo made Latexify to begin with.

ChrisRackauckas:
Both are deprecated and have their own Latexify recipes so these not only do nothing, but also throw warnings while doing nothing.

korsbo:
This looks fine to me. Thanks 🙂

korsbo:
aah, right. There's a mismatch between what triggers require (DiffEqBase) and what then gets loaded (ParameterizedFunctions). Should be easy enough to fix. 

I still don't mind dropping support for these pretty ancient packages though, so we can let this one stand.

PR #263, opened 2023-04-27
Convert extensions to recipes

gustaphe:
I've wanted to do this for a while, and it recently became much easier. Low priority, but maybe at least worth keeping in mind if we ever need to change either extension.

If this breaks something, that something was not in the tests.

### DIFF
--- a/ext/DataFramesExt.jl
+++ b/ext/DataFramesExt.jl
@@ -3,35 +3,13 @@ module DataFramesExt
 using Latexify
 isdefined(Base, :get_extension) ? (using DataFrames) : (using ..DataFrames)
 
-##################################################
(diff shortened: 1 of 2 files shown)


rodrigomha:
@gustaphe A quick question here. What is now the intended form to use create a markdown table from a Dataframe now that `mdtable(DataFrame)` is not supported directly? This broke some of my documentation and required to fix Latexify in 0.16.3 for now.

Thanks

gustaphe:
That is unfortunate, we should have had a test for that. 
I'm not currently at my computer, but `latexify(df)` or `latexify(df; env=:mdtable)` should do the same thing.

Issue #272, opened 2023-09-15
latexify does not support objects of type InlineStrings.String7


a2k42:
I was trying to convert a csv file to a markdown table and ran into an error whenever I had strings in my file (obviously a very common scenario)

Minimal example:

```julia
df = DataFrame(CSV.File("file.csv"))
latexify(df)
```

There's a [discourse suggesting](https://discourse.julialang.org/t/new-string7-type-from-result-of-csv-file-dataframe-makes-my-function-failing-which-takes-a-string-as-input/68985) that you need to handle `AbstractString` instead of just `String`. The breaking change happened a couple of years ago so maybe I've missed something?

> AssertionError: latexify does not support objects of type InlineStrings.String7.
    var"#_latexraw#45"(::Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:env,), Tuple{Symbol}}}, ::typeof(Latexify._latexraw), ::InlineStrings.String7)@latexraw.jl:109
    var"#process_latexify#156"(::Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:env,), Tuple{Symbol}}}, ::typeof(Latexify.process_latexify), ::InlineStrings.String7)@latexify_function.jl:49
    process_latexify@latexify_function.jl:40[inlined]
    var"#_latexinline#98"(::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, ::typeof(Latexify._latexinline), ::InlineStrings.String7)@latexinline.jl:4
    _latexinline(::InlineStrings.String7)@latexinline.jl:3
    _broadcast_getindex_evalf@broadcast.jl:683[inlined]
    _broadcast_getindex@broadcast.jl:656[inlined]
    getindex@broadcast.jl:610[inlined]
    copyto_nonleaf!(::Matrix{LaTeXStrings.LaTeXString}, ::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, typeof(Latexify._latexinline), Tuple{Base.Broadcast.Extruded{Matrix{Any}, Tuple{Bool, Bool}, Tuple{Int64, Int64}}}}, ::CartesianIndices{2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}, ::CartesianIndex{2}, ::Int64)@broadcast.jl:1068
[...]

PR #273, opened 2023-10-08
Bugfix/abstract string

a2k42:
Fixes https://github.com/korsbo/Latexify.jl/issues/272

### DIFF
--- a/docs/src/table_generator.jl
+++ b/docs/src/table_generator.jl
@@ -13,9 +13,9 @@ using Latexify
 struct KeywordArgument
     kw::Symbol
     env::Array{Symbol}
-    values::String
-    default::String
-    description::String
+    values::AbstractString
+    default::AbstractString
+    description::AbstractString
     types::Array{Symbol}
 end
 
--- a/docs/src/tutorials/inner_workings.md
+++ b/docs/src/tutorials/inner_workings.md
@@ -9,7 +9,7 @@ and
 
 - `latexoperation(ex::Expr, prevOp::AbstractArray)`
 
-These two methods are involved with all conversions to ``\LaTeX`` equations. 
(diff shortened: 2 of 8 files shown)


a2k42:
I'm still a bit of a Julia noob so this was quite a naive fix, but it seems to work. Fingers crossed it doesn't break anything!

gustaphe:
I think this is a bit overzealous, changing to `AbstractString` in some places where I can't imagine we would ever need one.

Could `@latexrecipe f(x::AbstractString) = string(x)` do the same job?

a2k42:
Using `@latexrecipe f(x::AbstractString) = string(x)` does also solve the `String7` error, but I run into another error.

When I fixed the problem using `AbstractString`, the error message given was quite instructive:

> in Latexify.jl:
> You are trying to create latex-maths from a `String` that cannot be parsed as
> an expression: `Hello World`.
> 
> `latexify` will, by default, try to parse any string inputs into expressions
> and this parsing has just failed.
> 
> If you are passing strings that you want returned verbatim as part of your input,
> try making them `LaTeXString`s first.
> 
> If you are trying to make a table with plain text, try passing the keyword
> argument `latex=false`. You should also ensure that you have chosen an output
> environment that is capable of displaying not-maths objects. Try for example
> `env=:table` for a latex table or `env=:mdtable` for a markdown table.

For my particular usecase `latexify(df, env=:mdtable, latex=false)` turned out to be the right thing to do. Adding `latex=false`, as I've just discovered, works just fine without any fixes.

I suspect I was trying to use `Latexify` in a way it wasn't supposed to be used, but only when resolving the String error did the correct usage become apparent.

There might be a more minimal change I could make that would make the "correct" error appear, but I'd need to do some digging so I understand the package usage better. Will close for now.

PR #277, opened 2024-01-10
Chaneg doc mentions of DiffEqBiological to Catalyst and update syntax

TorkelE:
A more throughout update can probably be made regarding this, but should probably wait until https://github.com/SciML/Catalyst.jl/issues/610 is fixed.

This simply changes mentioned of DiffEqBiological in the docs to Catalyst, and changes the various notations to use the latest version.


TorkelE:
Documentation seems busted. Might have to do with recent Julia updates or Documenter updates, see https://github.com/korsbo/Latexify.jl/pull/278 which provides very minimal changes.

korsbo (approved):

PR #285, opened 2024-03-30
fix nightly

t-bltg:
Required for Plots on `nightly` (fix `Iterators.Stateful` iteration), and test upcoming `1.11`.

Building docs were already broken before, left for another PR.

cc @korsbo.


korsbo:
Thanks, and sorry about the slow merge!

Issue #284, opened 2024-03-06
3D arrays/tensors


freemin7:
```
W = [
0 0.5 0 0 0 0 0 0 
0.5 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0;;;
0 0 0 0 0 0 0 0 
0 0 0 0 0 0 0 0
0 0 0 0.5 0 0 0 0
0 0 0.5 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0;;;
0 0 0 0 0 0 0 0 
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0.5 0 0
0 0 0 0 0.5 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0;;;
0 0 0 0 0 0 0 0 
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0.5
0 0 0 0 0 0 0.5 0
]

latexify(W) crashes with:
```
ERROR: BoundsError: attempt to access 8×8×4 Array{String, 3} at index [1, 1]
Stacktrace:
  [1] getindex(::Array{String, 3}, ::Int64, ::Int64)
    @ Base ./essentials.jl:14
  [2] _latexarray(arr::Array{Float64, 3}; adjustment::Symbol, transpose::Bool, double_linebreak::Bool, starred::Bool, arraystyle::Symbol, kwargs::Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:env,), Tuple{Symbol}}})
    @ Latexify ~/.julia/packages/Latexify/rf2cZ/src/latexarray.jl:49
  [3] _latexarray
    @ ~/.julia/packages/Latexify/rf2cZ/src/latexarray.jl:17 [inlined]
  [4] #_latexraw#46
    @ ~/.julia/packages/Latexify/rf2cZ/src/latexraw.jl:112 [inlined]
[...]

PR #286, opened 2024-04-14
Disallow tensors

gustaphe:
Solves #284 

Higher-dimensional tensors cannot be simply printed, so this PR throws an error. A bit more informative than the indexing error it would have caused before.


korsbo (changes requested):
Looks good, thanks, just fix the test on 1.6 and we'd be good to go

PR #287, opened 2024-04-15
Fix docs errors

devmotion:
The PR resolves the errors of the docs builds.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -26,6 +26,8 @@ makedocs(
         "tutorials/inner_workings.md",
     ],
     doctest = false,
+    checkdocs = :exports,
+    warnonly = :missing_docs
 )
 
 deploydocs(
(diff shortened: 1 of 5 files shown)


docs/make.jl
```diff
@@ -26,6 +26,8 @@ makedocs(
         "tutorials/inner_workings.md",
     ],
     doctest = false,
+    checkdocs = :exports,
+    warnonly = :missing_docs
```
devmotion, line 30: There are a few missing docs:
```julia
┌ Warning: 8 docstrings not included in the manual:
│ 
│   Latexify.render :: Tuple{LaTeXString}
│   Latexify.set_default :: Tuple{}
│   Latexify.@latexrun :: Tuple{Any, Vararg{Any}}
│   Latexify.reset_default :: Tuple{}
│   Latexify.get_default
│   Latexify.@latexify :: Tuple{Any, Vararg{Any}}
│   Latexify.@latexdefine :: Tuple{Any, Vararg{Any}}
│   Latexify.mdtable
│ 
│ These are docstrings in the checked modules (configured with the modules keyword) that are not included in canonical @docs or @autodocs blocks.
```
I was not sure where they should be put, so for now I just turned the errors into warnings.

PR #292, opened 2024-05-26
Allow recipes to set previous operation

gustaphe:
Previously, recipes had no way of affecting the previous operation. With this (somewhat ugly) change, a type that fundamentally represents for instance a multiplication or sum can tell latexify, so that things like parentheses work out right.

Part of fixing #280 

## TODO
* [x] Documentation
* [ ] Maybe more tests

### DIFF
--- a/docs/src/tutorials/recipes.md
+++ b/docs/src/tutorials/recipes.md
@@ -79,3 +79,31 @@ The use of `@latexrecipe` to redefine how an already supported type should be in
 
 
 If a recipe is defined within a module, everything should just work without the need to export anything. 
+
+The special keyword argument `operation` lets you specify that a type corresponds to a specific arithmetic operation.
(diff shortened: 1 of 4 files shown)


gustaphe:
## Without this,
```julia
julia> struct SimpleLength
x
end

julia> @latexrecipe function f(x::SimpleLength)
return Expr(:latexifymerge, x.x, "\\;\\mathrm{m}")
end

julia> @latexify $(SimpleLength(3))^2
```
```math
3\;\mathrm{m}^2
```

## With this
```julia
julia> @latexrecipe function f(x::SimpleLength)
operation := :*
return Expr(:latexifymerge, x.x, "\\;\\mathrm{m}")
end

julia> @latexify $(SimpleLength(3))^2
```
```math
\left( 3\;\mathrm{m} \right)^{2}
```

Issue #246, opened 2023-02-01
Missing infix operators for mathematics


joshniemela:
`latexify("a ∈ b")` results in `L"$\in\left( a, b \right)$"` and should probably return `L"$a \in b$"`.
`\forall` and `\exists` and probably other infix operators result in a crash.

gustaphe:
Neither `\forall` nor `\exists` are infix in julia:
```
julia> Meta.isbinaryoperator.((:∀, :∃))
(false, false)
```
, but `\forall` could be implemented. Do you think `\exists` should be treated as infix? I can't recall ever seeing it used as such. And do you have other ideas for operators that should be infixed?

joshniemela:
Yep you're right, exists isn't supposed to be infix but the others should be, I don't really have any operators in mind at the moment, maybe a lot of the logical operators if they don't already infix themselves.

jakobjpeters:
My package wants to handle infix operators too, which is handled incorrectly.

jakobjpeters:
I discovered that you need to wrap the string in a `Symbol`. This should be documented though.
```julia
julia> latexify(Symbol("a ∈ b"))
L"$a \in b$"
```

gustaphe:
Could I ask you guys to try out #296 ? It's difficult to test if this works right, since it is partially taste.
Specifically, I want to know if parentheses work as you expect, and if there are any operators you miss.

PR #296, opened 2024-07-04
Infix operators and paranthesis precedence

### LINKED ISSUE #168: Unnecessary parentheses from `:(-x*y)`
`-x` does not need parentheses in this case.

### LINKED ISSUE #239: Why are left and right brackets automatically added when executing the latexify("\alpha>\beta")??
I used the code below.

using LaTeXStrings, Latexify
latexify("α > β")  

REPL shows the wrong answer : L"$\left( \alpha > \beta \right)$"
What should I do to get correct answer,  L"$ \alpha > \beta $"  ???

### LINKED ISSUE #279: Request: support ≤ and ≥ the same as <= and >=
This looks great!
image

However, this does not.
image

Would it be possible to support the ≤ and ≥ Unicode characters, so both my code and LaTeX look good?

gustaphe:
I headed out to add some infix comparison operators, but wasn't happy to have to defend keeping the weird parenthesis issues, so I fixed that too.

## List of infix comparison operators
These are the ones currently in the comparison dictionary:
image

Closes #168 
Closes #239
Closes #279
Does much of #246


gustaphe:
Will  probably need to forgo "\forall", it's not valid in 1.6

PR #300, opened 2024-08-06
Separate aligned equations correctly

hersle:
Aligned equations were separated by `=&`, which produces an unintended space (https://github.com/SciML/ModelingToolkit.jl/issues/2894) around the equality sign. This changes them to be separated by `&=` instead, which is the most common way to do it and removes the unindented space. See also [this StackExchange post](https://tex.stackexchange.com/questions/41074/relation-spacing-error-using-in-aligned-equations).

Does this have any unintended side effects I am not thinking about?


gustaphe:
Looks uncontroversial to me.
Thanks!

Issue #78, opened 2019-10-15
Find a less restrictive way of testing?


korsbo:
Automatic testing of latexify is currently a little painful since it relies on precise equality between the string generated by a function call and a string typed out in the test script. Generating such tests is pretty easy using the `Latexify.@generate_test` macro, but this exact string matching gives no leeway for making otherwise inconsequential changes to how `latexify` formats its output. 

Is there a way of testing that `latexify` gives the desired result without being completely inflexible when it comes to making small changes? 

This is related to the question of what should be considered part of the API, #77.

PR #303, opened 2024-08-08
WIP: Visual regression tests

gustaphe:
An idea to solve #78 , using VisualRegressionTests.jl
Not ready or even sure if it's that useful, but I wanted to see if I could get CI to collaborate.

Also needed an option to render without transparency, which should probably be implemented either way.


gustaphe:
There we go. We could probably come back to this before we make any big changes (like #161)

PR #305, opened 2024-08-27
render cbrt and fourthroot functions

mxhbl:
Latexify currently only converts cube roots or fourth roots if they are written using `∛ ` or `∜`, but not if they are written with the `cbrt` or `fourthroot` functions.
This PR fixes this.


gustaphe:
Perfect, thanks!

PR #308, opened 2024-09-15
Latexify undefined array elements

gustaphe:


gustaphe:
```julia
julia> undefarr = Array{Any,2}(undef, 2, 2)
2×2 Matrix{Any}:
 #undef  #undef
 #undef  #undef

julia> undefarr[1] = "x"
"x"

julia> render(latexify(undefarr))
```
image

ChrisRackauckas:
That looks reasonable to me!

damianodegaspari:
If I am not mistaken, \cdot is the same symbol used for sparse arrays. In that context it means 0 instead of undefined. Personally, I don't think it's a problem: context should make it clear enough. In my specific case, as an inexperienced person just trying to initialize a matrix, I would have been happy with any representation which didn't look like an error. Thus I'm fine with this solution (as I said, I would have been fine with basically everything).

PR #309, opened 2024-09-17
Allow unicode cdot to be considered operator alongside *

auriium2:
Sometimes you want to assign unicode cdot to dot product, etc but latexify will render it like a function. This PR should allow the unicode cdot to be rendered like the asterisk. It is also immune to kwargs cdot = false because sometimes you want to notate the dot in certain circumstance, etc

### DIFF
--- a/src/latexoperation.jl
+++ b/src/latexoperation.jl
@@ -48,7 +48,15 @@ function latexoperation(ex::Expr, prevOp::AbstractArray; kwargs...)::String
             i == length(args) || (str *= cdot ? " \\cdot " : " ")
         end
         return str
-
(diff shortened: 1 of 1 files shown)


gustaphe:
Nice idea! I think it would be better for maintainability to add `:⋅ => "\\cdot"` to `binary_operators` in `symbol_translations.jl`, then the rest should follow automatically. 
I should probably document this somewhere.

Could you write some tests? Especially try to catch the "is not affected by cdot=false" behavior, and that it's parenthesized correctly.

superyuuki:
That's a good suggestion. I'll try to write some tests as well.

Issue #307, opened 2024-09-14
Rendering higher order tensors and uninitialized arrays.


damianodegaspari:
Hi,

I am opening this issue as suggest to me in the Julia discord topic [Matrices and tensors with Symbolics.jl](https://discourse.julialang.org/t/matrices-and-tensors-with-symbolics-jl/119394).  It is about two problems I encountered while using `Symbolics.jl` with `Pluto.jl`. I rewrite everything here.

2) *Pluto gives warning (or is it an error?) when creating Arrays indexed by more than two elements.*

Behaviour in the REPL is what I would expect (final output cut for legibility):

```
julia> using Symbolics

julia> Symbolics.variables(:H, 1:2, 0:3, 0:3, 0:3)
2×4×4×4 Array{Num, 4}:
[:, :, 1, 1] =
 H₁ˏ₀ˏ₀ˏ₀  H₁ˏ₁ˏ₀ˏ₀  H₁ˏ₂ˏ₀ˏ₀  H₁ˏ₃ˏ₀ˏ₀
 H₂ˏ₀ˏ₀ˏ₀  H₂ˏ₁ˏ₀ˏ₀  H₂ˏ₂ˏ₀ˏ₀  H₂ˏ₃ˏ₀ˏ₀

[:, :, 2, 1] =
 H₁ˏ₀ˏ₁ˏ₀  H₁ˏ₁ˏ₁ˏ₀  H₁ˏ₂ˏ₁ˏ₀  H₁ˏ₃ˏ₁ˏ₀
 H₂ˏ₀ˏ₁ˏ₀  H₂ˏ₁ˏ₁ˏ₀  H₂ˏ₂ˏ₁ˏ₀  H₂ˏ₃ˏ₁ˏ₀

[:, :, 3, 1] =
 H₁ˏ₀ˏ₂ˏ₀  H₁ˏ₁ˏ₂ˏ₀  H₁ˏ₂ˏ₂ˏ₀  H₁ˏ₃ˏ₂ˏ₀
 H₂ˏ₀ˏ₂ˏ₀  H₂ˏ₁ˏ₂ˏ₀  H₂ˏ₂ˏ₂ˏ₀  H₂ˏ₃ˏ₂ˏ₀

[:, :, 4, 1] =
[...]

PR #310, opened 2024-09-18
Add exception types, and a sentinel keyword argument

gustaphe:
Discussion material for #307 
(I may have gone overboard with the error types)

Two concepts to solve the problem of `Pluto+Symbolics` breaking. Both requires a little intervention at the call site:

1. Use `try/catch` to guard for `LatexifyException`s, and return something else completely
2. Use the `sentinel` keyword to make `latexify` return *something*

I think I prefer 2, it means everything but the offending type, for instance, gets latexified sensibly. So
```
julia> struct HasNoRecipe end
julia> latexify([1, "x", HasNoRecipe()]; sentinel="\\mathrm{NA}")
L"\begin{equation}
\left[
\begin{array}{c}
1 \\
x \\
\mathrm{NA} \\
\end{array}
\right]
\end{equation}
"
```
Not married to any of it, nor the name "sentinel".


gustaphe:
Do not merge, made a rebasing mistake, will make a new PR.

PR #323, opened 2024-12-13
Use thin \, space for multiplication when cdot = false

hersle:
Problem: currently `latexify("A*B"; cdot=false)` and `latexify("AB")` are visually indistinguishable, because `$A B$` and `$AB$` renders the same in Latex as $AB$.

Solution: instead of a blank space, use the thin space `\,` as a "multiplication operator" when `cdot = false`. Then `latexify("A*B"; cdot=false)` gives `$A \, B`, i.e. $A \\, B$, which looks slightly different from $AB$. I think this is slightly better default behavior.


gustaphe:
Thank you for the contribution!

I think this leads to incorrect LaTeX code, there's a reason the space character is ignored.
Perhaps a keyword argument `mult_symbol = "\\,"` would make this configurable without breaking the default.
`cdot=false` would be an override for `mult_symbol = ""`.

hersle:
I think I see what you mean, and I agree that `mult_symbol` would be an even better and more flexible solution.

Having both `cdot` and `mult_symbol` is superfluous, though, and could unnecessarily clutter the available options. I suggest to replace `cdot` with `mult_symbol`. What do you think about letting `cdot=false` be an override for `mult_symbol=""`, as you suggested, but with a deprecation warning that says how one should use `mult_symbol` instead?

gustaphe:
I think the deprecation warning is a good solution. There is no current plan for a breaking release, but if `Base.depwarn` is used it will be painless to users to keep the duality indefinitely. 

I feel like `mult_symbol` should take precedence, so `mult_symbol="\\cdot", cdot=false` should boil down to `mult_symbol="\\cdot"` and vice versa.

Would you mind writing this up?

hersle:
Yes, I agree. Thanks for the input! I hope to get around to do it in the near future.

PR #325, opened 2024-12-18
Replace boolean cdot keyword argument in favor of more flexible mult_symbol string

hersle:
Follow-up to https://github.com/korsbo/Latexify.jl/pull/323; see discussion there.

@gustaphe I figured the simplest way of handling the deprecation of `cdot` in code is to convert it to the equivalent `mult_symbol`. This means `cdot` takes precedence over `mult_symbol` in the event that they are both set, which is the opposite of what we agreed. I can argue both ways; if you think it is more natural for `mult_symbol` to take priority, or there is anything else, I am happy to update it.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -156,7 +156,7 @@ Note that this changes Latexify.jl from within and should therefore only be used
 
 The calls are additive so that a new call with 
 ```julia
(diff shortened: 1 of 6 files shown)


gustaphe:
This is very nice. Thanks!

PR #326, opened 2025-02-03
Parse all four levels of \dot

hersle:
Julia seems to have up to four possible levels of \dot-ting.
I encountered errors when rendering some of these that were not implemented.


gustaphe:
Do you mean that this was an error before this implementation, or does some engine (like MathJax) not render `\dddot` correctly?

Either way, LGTM. It's generally fine for Latexify to make output that some engines might not recognize in edge cases, as long as it is valid LaTeX.

hersle:
Before the implementation, in VSCode:
```julia
render(latexify("τ⃛")) # \tau\dddot
```
gives
image

Issue #339, opened 2025-08-08
`render` does not provide a way to set LaTeX or Ghostscript flags


Ickaser:
To help generate documentation programmatically for UnitfulLatexify (https://github.com/JuliaPhysics/Unitful.jl/pull/795), I want to call `Latexify.render` with `tectonic_jll`. However, for this case, we have a large number of units not defined in siunitx, which on LaTeX compilation are undefined control sequences. I want these undefined units to simply not be printed in the final document, and fortunately there is a command line flag for `tectonic` which gives this behavior: `tectonic -Z continue-on-errors ...`

I would be happy to submit a PR adding this flag to the `TectonicExt` of Latexify but I suspect in general that users will not want that behavior. Would it be best to add a custom keyword argument for this? Or is there an idiomatic Julia way to pass general keyword arguments through a function to a shell call?

Here is the key method:
https://github.com/korsbo/Latexify.jl/blob/d416db96652d09ab97533c576f2023ed1df83f19/ext/TectonicExt.jl#L7-L10

If this is a recursive call from `render(s, MIME("image/png"))`, we would like the arguments to pass through that call, so it's not enough to add a kwarg to this method, it is also needed here:

https://github.com/korsbo/Latexify.jl/blob/d416db96652d09ab97533c576f2023ed1df83f19/src/utils.jl#L148-L185

PR #340, opened 2025-08-08
Add kwargs for flags for all command line calls

Ickaser:
This is a rudimentary attempt at dealing with #339 so that I can improve docs in https://github.com/JuliaPhysics/Unitful.jl/pull/795 so that I can eventually fix https://github.com/JuliaPlots/Plots.jl/issues/5093.

Basic idea is that `render` gets some new kwargs: `ghostscript_flags`, `dvipng_flags`, `tectonic_flags`, `lualatex_flags`, `dvilualatex_flags`. Since there is already a slurped `kw...` in all of the methods, I don't think this is too disruptive.

`ghostscript_flags` and `dvipng_flags` have default values, related to rastering. All of the other flag kwargs default to empty strings.

### DIFF
--- a/ext/TectonicExt.jl
+++ b/ext/TectonicExt.jl
@@ -4,8 +4,8 @@ import Latexify.render, Latexify._compile
 isdefined(Base, :get_extension) ? (using tectonic_jll) : (using ..tectonic_jll)
 __precompile__(false)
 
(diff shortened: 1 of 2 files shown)


gustaphe:
No, I think this is good. 

I've enjoyed following your crusade, you're getting there!

